PR #5, opened 2019-11-12
Add default role definitions for more apps

astrozzc:
This PR adds default role definitions for the following applications:

* ansible-automation
* ansible-hub
* remediations
* subscriptions

Each of these applications needs a base, global read/write permission for
itself, as well as inventory, which is a dependent service which will be
receiving permissions from the upstream service. These permissions are set on
a base role for each application.

### DIFF
--- a/configs/ansible-automation.json
+++ b/configs/ansible-automation.json
@@ -1,0 +1,18 @@
+{
+  "roles": [
+    {
+      "name": "Ansible-automation Access",
+      "description": "An all access role which grants read and write permissions.",
+      "system": true,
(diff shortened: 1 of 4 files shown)


coderbydesign:
Hey @astrozzc! Based on the conversation this morning, it sounds like Nicole will be verifying which default roles are required for which apps?

astrozzc:
> Hey @astrozzc! Based on the conversation this morning, it sounds like Nicole will be verifying which default roles are required for which apps?

OK. Let's wait for that and then make changes 👍

PR #12, opened 2020-01-07
Create approval user role and use it as platform default

bzwei:
Create `Approval User` role with restricted permissions and turn on the `platform_defaut` flag.

### DIFF
--- a/configs/approval.json
+++ b/configs/approval.json
@@ -1,6 +1,24 @@
 {
   "roles": [
     {
+      "name": "Approval User",
+      "description": "An approval user role which grants permissions to read/create a request.",
+      "system": true,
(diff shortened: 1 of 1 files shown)


configs/approval.json
```diff
@@ -1,5 +1,23 @@
...
+      "description": "An approval user role which grants permissions to read/create a request.",
+      "system": true,
+      "platform_default": true,
+      "version": 1,
+      "access": [
+        {
+          "permission": "approval:requests:read"
+        },
+        {
+          "permission": "approval:requests.create"
```
coderbydesign: Is this supposed to be `approval:requests:create`?
```suggestion
          "permission": "approval:requests:create"
```

PR #18, opened 2020-01-30
Add a default_platform role for migration-analytics

coderbydesign:


coderbydesign:
@astrozzc correct. Inventory isn't even rolling out with RBAC lite, so we don't really need to worry about those at all right now.

PR #20, opened 2020-02-07
RHCLOUD-3676 default role for Remediations

jharting:
based on https://docs.google.com/document/d/14hTY9K5LKV3INlFfjbmiW3LgGPp8_-Xb1zIQ6R0M6CA/edit

### DIFF
--- a/configs/remediations.json
+++ b/configs/remediations.json
@@ -1,0 +1,22 @@
+{
+  "roles": [
+    {
+      "name": "Remediations Access",
+      "description": "A role which grants read and write permissions.",
+      "system": true,
+      "platform_default": true,
+      "version": 2,
+      "access": [
+        {
+          "permission": "remediations:remediation:read"
+        },
+        {
+          "permission": "remediations:remediation:write"
+        },
+        {
+          "permission": "inventory:*:*"
+        }
+      ]
+    }
+  ]
+}


configs/remediations.json
```diff
@@ -0,0 +1,22 @@
+{
+  "roles": [
+    {
+      "name": "Remediations Access",
+      "description": "A role which grants read and write permissions.",
+      "system": true,
+      "platform_default": true,
+      "version": 4,
```
coderbydesign: It doesn't _really_ matter, but this doesn't need to be 4, it could be 2. The seeding versions technically start at 2 and bump up from there. Again, it doesn't make a huge difference, as you'd just be incrementing from 4 instead of 2 if you need to make changes to this role in the future. We don't store the version, just check for a diff.

configs/remediations.json
```diff
@@ -0,0 +1,22 @@
...
+      "access": [
+        {
+          "permission": "remediations:remediation:read"
+        },
+        {
+          "permission": "remediations:remediation:write"
+        },
+        {
+          "permission": "inventory:*:*"
```
coderbydesign: Will remediations be making calls to inventory to need this implicit permission? Inventory does not yet implement RBAC, but this would account for it when they do.

jharting: Yes, it's making calls to inventory to resolve host info based on host identifier.

configs/remediations.json
jharting: Oh I see. I saw other roles using 4 so I did the same but I'll change this to 2.

jharting:
@coderbydesign I updated the version based on your comment.

PR #24, opened 2020-02-26
Added config for patchman

semtexzv:
App name is `patch` to match default our url path on the website & the api.
Current implementation calls inventory fr retrieving system profiles.

### DIFF
--- a/configs/patch.json
+++ b/configs/patch.json
@@ -1,0 +1,16 @@
+{
+  "roles": [
+    {
+      "name": "Patchman Access",
+      "description": "An all access role which grants read and write permissions.",
+      "system": true,
+      "platform_default": true,
+      "version": 2,
+      "access": [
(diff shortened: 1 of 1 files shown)


configs/patch.json
```diff
@@ -0,0 +1,19 @@
...
+      "description": "An all access role which grants read and write permissions.",
+      "system": true,
+      "platform_default": true,
+      "version": 2,
+      "access": [
+        {
+          "permission": "patch:*:*"
+        },
+        {
+          "permission": "inventory:*:*"
```
coderbydesign: Hey @semtexzv I just realized, we have [a ticket to remove all of the implicit inventory permissions.](https://projects.engineering.redhat.com/browse/RHCLOUD-4775) 

Could you pull that permission out? Then we should be good. 😄

PR #37, opened 2020-03-18
[WIP] Remove resourceDefinitions and add approve and admin permissions

bzwei:
version 4:
Remove `resourceDefinitions`.
Add `approve` permission to approver role.
Add `admin` permission to administrator role.

### DIFF
--- a/configs/approval.json
+++ b/configs/approval.json
@@ -2,10 +2,10 @@
   "roles": [
     {
       "name": "Approval User",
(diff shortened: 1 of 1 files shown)


bzwei:
cc @hsong-rh

gmcculloug:
@bzwei Marking as WIP so we can discuss this approach.  Discussing how to handle the admin definition for Catalog with @mkanoor now.

PR #38, opened 2020-03-19
Updated Approval service permissions

hsong-rh:
Updated role permissions for Approval service.

### DIFF
--- a/configs/approval.json
+++ b/configs/approval.json
@@ -1,27 +1,9 @@
 {
   "roles": [
     {
-      "name": "Approval User",
-      "description": "An approval user role which grants permissions to read/create a request.",
-      "system": true,
-      "platform_default": true,
-      "version": 2,
-      "access": [
-        {
-          "permission": "approval:requests:read"
-        },
-        {
-          "permission": "approval:requests:create"
-        },
-        {
-          "permission": "approval:actions:create"
-        }
-      ]
-    },
-    {
       "name": "Approval Administrator",
       "system": true,
-      "version": 3,
+      "version": 4,
       "description": "An Approval administrator role that grants create, read, update and destroy permissions",
       "access": [
         {
@@ -29,9 +11,9 @@
           "resourceDefinitions": [
             {
               "attributeFilter": {
-                "value": "*",
-                "key": "id",
-                "operation": "equal"
+                "key": "scope",
+                "operation": "equal",
+                "value": "admin"
               }
             }
           ]
@@ -41,9 +23,9 @@
           "resourceDefinitions": [
             {
               "attributeFilter": {
-                "value": "*",
-                "key": "id",
-                "operation": "equal"
+                "key": "scope",
+                "operation": "equal",
+                "value": "admin"
               }
             }
           ]
@@ -53,9 +35,9 @@
           "resourceDefinitions": [
             {
               "attributeFilter": {
-                "value": "*",
-                "key": "id",
-                "operation": "equal"
+                "key": "scope",
+                "operation": "equal",
+                "value": "admin"
               }
             }
           ]
@@ -65,9 +47,9 @@
           "resourceDefinitions": [
             {
               "attributeFilter": {
-                "value": "*",
-                "key": "id",
-                "operation": "equal"
+                "key": "scope",
+                "operation": "equal",
+                "value": "admin"
               }
             }
           ]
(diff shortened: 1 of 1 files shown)


gmcculloug:
@mkanoor please review

configs/approval.json
```diff
@@ -113,44 +107,160 @@
...
+                "key": "scope",
+                "operation": "equal",
+                "value": "admin"
               }
             }
           ]
         },
         {
-          "permission": "approval:actions:create",
+          "permission": "approval:taglinks:create",
```
mkanoor: @hsong-rh are you exposing the tag links or are you going to control it from workflows?

hsong-rh: @mkanoor Changed to control from workflows

configs/approval.json
```diff
@@ -113,9 +131,41 @@
...
+            }
+          ]
+        }
+      ]
+    },
+    {
+      "name": "Approval User",
+      "system": true,
+      "version": 4,
+      "description": "An approval user role which grants permissions to create/read/cancel a request.",
```
bzwei: `.... /cancel a request, and read workflows." `

Please add read action permission.

configs/approval.json
```diff
@@ -113,9 +131,41 @@
...
+                "key": "scope",
+                "operation": "equal",
+                "value": "user"
+              }
+            }
+          ]
+        },
+        {
+          "permission": "approval:requests:create",
+          "resourceDefinitions": [
```
mkanoor: @hsong-rh Do we need resourceDefintiion when creating a resource. I am thinking that the scope comes when we are dealing with existing records in the DB. So it would be needed for read, update, delete but not for create. In the RBAC permission check to see if a user has permission to create you would be doing
```
x = Insights::API::Common::RBAC::Access.new().process
x.accessible?('requests','create')
```

configs/approval.json
```diff
@@ -125,9 +175,33 @@
...
+          ]
+        },
+        {
+          "permission": "approval:workflows:read",
+          "resourceDefinitions": [
+            {
+              "attributeFilter": {
+                "key": "scope",
+                "operation": "equal",
+                "value": "user"
```
mkanoor: @hsong-rh Are the workflows created per user. This permission is saying that the logged in user would only be able to see the workflows he or she created. He/she would not be able to see all workflows in the tenant. If he/she is allowed to read all Workflows then the scope should be "admin"

hsong-rh: @mkanoor Approval service actually does not use `user scope`. The better way is only having permission here, no need to defined `resourceDef`, like we mentioned before.

configs/approval.json
hsong-rh: No need to define `resourceDefinition` for regular user, only permissions are enough.

configs/approval.json
```diff
@@ -1,133 +1,162 @@
...
+          "permission": "approval:actions:read",
           "resourceDefinitions": [
             {
               "attributeFilter": {
-                "value": "*",
-                "key": "id",
-                "operation": "equal"
+                "key": "scope",
+                "operation": "equal",
+                "value": "admin"
```
bzwei: value is user

configs/approval.json
```diff
@@ -1,133 +1,162 @@
 {
   "roles": [
     {
-      "name": "Approval User",
-      "description": "An approval user role which grants permissions to read/create a request.",
+      "name": "Approval Administrator",
       "system": true,
-      "platform_default": true,
-      "version": 2,
+      "version": 4,
```
mkanoor: @hsong-rh Should this be version 3 and not 4.

hsong-rh:
@mkanoor Per discussed, I added back `resourceDefinitions` for `create` actions.

configs/approval.json
hsong-rh: Do we need to keep `version` in sync?

configs/approval.json
```diff
@@ -113,9 +131,41 @@
...
+                "operation": "equal",
+                "value": "admin"
+              }
+            }
+          ]
+        }
+      ]
+    },
+    {
+      "name": "Approval User",
```
mkanoor: @hsong-rh Are we missing
"platform_default": true,

For Approval User

bzwei: Good catch!

hsong-rh: Added! Thanks!

mkanoor:
approval:workflows:destroy =>  approval:workflows:delete

mkanoor:
@hsong-rh Do you need permissions for the template object?

mkanoor:
👍

bzwei:
@mkanoor only admin can read templates. They are created internally.

PR #40, opened 2020-03-24
Added support for scopes

mkanoor:
We have added support for 
 - admin 
 - group
 - user 

scopes to differentiate permission sets for different roles.

### DIFF
--- a/configs/catalog.json
+++ b/configs/catalog.json
@@ -3,98 +3,341 @@
     {
       "name": "Catalog Administrator",
       "system": true,
-      "version": 4,
+      "version": 5,
       "description": "A catalog administrator roles grants create,read,update, delete and order permissions",
       "access": [
         {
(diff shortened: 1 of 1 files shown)


mkanoor:
@gmcculloug 
Please review

configs/catalog.json
```diff
@@ -3,113 +3,350 @@
...
+              "attributeFilter": {
+                "key": "scope",
+                "operation": "equal",
+                "value": "admin"
+              }
+            }
+          ]
         },
         {
           "permission": "catalog:portfolios:create"
```
gmcculloug: @mkanoor why was this one changed?

mkanoor: @gmcculloug When you are doing a Create on a Portfolio, there are no specific instances involved like when you do a read, update, delete. So the scope is not needed. The API is just going to be a permission check without the scope.

PR #49, opened 2020-04-03
Add system role for remediations

astrozzc:
Remediations have created this role for some tenant.
Those ones have default version 1. Here the version is
set to 2, which will trigger the update when seeding.

### DIFF
--- a/configs/remediations.json
+++ b/configs/remediations.json
@@ -1,6 +1,18 @@
 {
   "roles": [
     {
+      "name": "Remediations administrator",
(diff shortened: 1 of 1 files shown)


astrozzc:
From the seed_role function, the update of the custom role seems won't change the role name and role id, which might not break the current reference to the role.

astrozzc:
Just confirmed that Remediations team wants the "Remediations user" to be the platform_default role, not the "administrator".

PR #55, opened 2020-04-08
Added new permission catalog:tenant:update

mkanoor:
This allows us to do Tenant Settings and Seeding on
the Catalog Side. Currently those actions are checking
for Catalog Administrator Role.

### DIFF
--- a/configs/catalog.json
+++ b/configs/catalog.json
@@ -238,14 +238,17 @@
           ]
         },
         {
-          "permission": "catalog:tenant:update"
+          "permission": "catalog:tenants:update"
+        },
+        {
+          "permission": "catalog:tenants:read"
         }
       ]
     },
     {
       "name": "Catalog User",
       "system": true,
-      "version": 7,
+      "version": 8,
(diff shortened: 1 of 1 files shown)


mkanoor:
@gmcculloug Please review

configs/catalog.json
```diff
@@ -236,6 +236,9 @@
               }
             }
           ]
+        },
+        {
+          "permission": "catalog:tenant:update"
```
gmcculloug, line 241: @mkanoor Does it make sense to include an `attributeFilter` for this permission, like the admin scopes on the others in this role?

mkanoor: @gmcculloug 
Since we are using acts_as_tenant gem we are limited to working with a single tenant at a time. We were using the scopes when there were multiple instances and we need to split the instances based on admin, user and group. Tenants are created automatically in our database so we dont need the create and neither do we need the delete. I thought about catalog:tenant:read if we need to read the tenant settings for admin and user.

PR #66, opened 2020-06-04
Add available permissions

astrozzc:
For each app, define the available permissions, which could be selected
when creating custom roles.

It is highly recommended to also add the new permissions in the canned role
to these file.


README.md
```diff
@@ -23,31 +23,49 @@ with the json files in the configs folder of rbac-config.
...
-
 Set the version to 2 for the new role in order to trigger the seeding in the rbac service.
-
-Format of permissions
-~~~~~~~~~~~~~~~~~~~~~
-The permission should have the format of <application>:<resource>:<action>. There are no restrictions on the
+If you have a new permission added, don't forget to add it to the permissions folder too. See "Add new permissions"
+section below for more details.
+~~~~~~~~~~~~~~~~~~~~~~
+Format of permissions:
```
coderbydesign, line 38: I think this is causing the formatting to be off. It's displaying odd for me.
Screen Shot 2020-06-09 at 11 13 17 AM

README.md
```diff
@@ -23,31 +23,49 @@ with the json files in the configs folder of rbac-config.
...
 If you update the name of the role, it will generate a new role but keep the old one. Please reach out to 
 RBAC team to resolve this.
 
 Delete roles
--------------
+
 Currently, if you delete a role from the config, it will still exist. Please reach out to RBAC team to delete them from database.
 
 
+Add new permissions
```
coderbydesign, line 55: I think it might be good to have a README.md in each of the directories for /permissions/ and /configs/ with a sample of what they should look like?

coderbydesign:
@astrozzc I'm wondering, would it make sense to have /configs/permissions and /configs/roles?

README.md
astrozzc: Oh, I deliberately make it look like this. First level is Add new/Update/Delete roles. Then the "Format of permissions" is a kind of high light. 

Are you referring to this high light? Do you want me to make it same level as the "Add new/Update/Delete roles"?

README.md
astrozzc: Good idea, I will add a README.md in those directories. Only samples in the README, right?

astrozzc:
> @astrozzc I'm wondering, would it make sense to have /configs/permissions and /configs/roles?

Yea, that would be better. I was thinking not affecting the current usage so I put the permissions in another directory. But I agree, I will update the structure in the e2e.

README.md
astrozzc: Turns out the create configmap will use the README file, therefore it could not be added inside the directories. But I added a line to let them refer to existing files for samples.

README.md
astrozzc: Updated, I put them to the same level with the others 😉

configs/permissions/complicance.json
```diff
@@ -0,0 +1,3 @@
+{
```
catastrophe-brandon, line 1: We should probably address the typo in the file name. complicance.json -> compliance.

astrozzc: Nice catch 👍

PR #68, opened 2020-06-11
Add permissions for OrderProcess

hsong-rh:
Added permissions for newly introduced `OrderProcess` model in https://github.com/RedHatInsights/catalog-api/pull/761

### DIFF
--- a/configs/catalog.json
+++ b/configs/catalog.json
@@ -280,7 +280,7 @@
           ]
         },
         {
-          "permission": "catalog:order_processes:write",
+          "permission": "catalog:order_processes:update",
           "resourceDefinitions": [
             {
               "attributeFilter": {
--- a/configs/catalog.json
+++ b/configs/catalog.json
@@ -3,7 +3,7 @@
     {
       "name": "Catalog Administrator",
       "system": true,
-      "version": 6,
+      "version": 7,
       "description": "A catalog administrator roles grants create,read,update, delete and order permissions",
       "access": [
         {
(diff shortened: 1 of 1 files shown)


configs/catalog.json
```diff
@@ -242,6 +242,54 @@
...
+              "attributeFilter": {
+                "key": "scope",
+                "operation": "equal",
+                "value": "admin"
+              }
+            }
+          ]
+        },
+        {
+          "permission": "catalog:order_processes:write",
```
gmcculloug, line 283: @mkanoor Should this be `write` or `update`?

I'm concerned I see both uses which is inconsistent and confusing.

gmcculloug:
@mkanoor @eclarizio Please review

configs/catalog.json
eclarizio: I think this is probably supposed to be `update`, since we're doing `access.scopes(table_name, verb)`, and we will likely be passing in `"update"` as the verb.

hsong-rh: I followed the pattern for `order_item` and `order`, thought it will be same for `order_process`. In approval, there is no `write` action.

hsong-rh: Changed to `update`. Then in which situation we should use `write`?

eclarizio: @hsong-rh Yeah, it's kinda weird. We don't currently check permissions against orders or order items. If you want to submit an order, the check that actually happens is "catalog:portfolio:order". We don't have any sort of permission check surrounding creating/updating/deleteing/listing orders or order items right now, so I think the "write" was just put there originally and since it was never used, it was never technically wrong.

gmcculloug: @mkanoor Please review and provide feedback.  I sense we want to be `using` update as part of the CRUD permissions we expose on the objects, unless `write` has a special meaning here.

mkanoor: @gmcculloug update is the new verb previously we used to have write but that didn't allows us to distinguish between create and update. So write is no longer used. Currently we have write for things that the system creates like progress messages, order items we could convert them to 2 separate items with create/update down the road. Any objects that is controlled by a logged in user is using CRUD.

mkanoor:
👍 looks good

configs/catalog.json
gmcculloug: @mkanoor Not for this PR, but would you recommend splitting the existing "write" permission into "create" and "update" permissions?

mkanoor: @gmcculloug Yes we should try and split the write to create/update for objects that get exposed. Since this OrderProcess object is going to get exposed it makes sense to split the write as create and update.

PR #74, opened 2020-07-20
Add links permissions to order process

hsong-rh:
https://projects.engineering.redhat.com/browse/SSP-1660

Add two permissions: `link` and `unlink` to order process for `admin` only.

### DIFF
--- a/configs/permissions/catalog.json
+++ b/configs/permissions/catalog.json
@@ -2,6 +2,7 @@
     "approval_requests": ["read", "write"],
     "orders": ["read", "write", "order"],
     "orders_items": ["read", "write", "order"],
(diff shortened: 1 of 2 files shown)


hsong-rh:
@coderbydesign Thank you, Keith. I do have question about this new added permission file: does it require to list out all resources, with all possible actions?

coderbydesign:
@hsong-rh - Not at the moment, no. This is the first iteration of custom role creation support, which is initially limited to cost, but this is an effort to move in that direction for all teams. Thanks! 😄

hsong-rh:
@coderbydesign I see, you can merge this, thanks.

PR #75, opened 2020-07-20
Add permissions and roles for Automation Analytics 

AparnaKarve:
Permissions and Roles for the Automation Analytics application.


astrozzc:
LGTM

PR #76, opened 2020-07-27
Bump Automation Analytics role versions

coderbydesign:
An issue with role seeding prevented the permissions from being attributed to the
new Automation Analytics roles from #75.

This bumps the version, to rerun access/permission creation for these roles based
on the new fix [1] in RBAC.

[1] https://github.com/RedHatInsights/insights-rbac/pull/317


coderbydesign:
cc: @AparnaKarve

PR #80, opened 2020-08-03
Inventory rbac v1

tahmidefaz:


configs/permissions/inventory.json
```diff
@@ -0,0 +1,3 @@
+{
+    "*": ["*", "read", "write"]
```
coderbydesign, line 2: To clarify, you'll want the following permissions available for custom role creation, but you'll only create two roles, one for admin and one for viewer, but won't be using the write permission explicitly out of the gate, correct?
```
inventory:*:*
inventory:*:read
inventory:*:write
```

tahmidefaz: Yes, I think that's what the team agreed on at this point. I'll update this PR if things change.

configs/permissions/inventory.json
```diff
@@ -0,0 +1,4 @@
+{
+    "*": ["*"],
+    "host_data": ["read", "write"]
```
jharting, line 3: \<nitpick\/\> I think that the `_data` part is implicit i.e. a user can have read access to host_data, remediation_data, advisory_data, subscription_data, ... or we can just say read access to host, remediations, advisory, subscription, ... . I think the latter is cleaner and more in line with other apps.

configs/roles/inventory.json
```diff
@@ -0,0 +1,16 @@
+{
+  "roles": [
+    {
+      "name": "Inventory Administrator",
+      "description": "An Inventory Administrator role that grants read and write permissions to all Inventory resources.",
+      "system": true,
+      "platform_default": true,
+      "version": 1,
```
coderbydesign, line 8: Due to an idiosyncrasy in the way our seeding works, this needs to be started at `"version": 2`.
```suggestion
      "version": 2,
```

tahmidefaz: Thanks! Done! :+1:

PR #81, opened 2020-08-17
Advisor Custom Permissions [RFR]

dkuc:


PaulWay:
I think this aligns well with Rob's vision of the permissions we want to allow users to have.

configs/permissions/insights.json
```diff
@@ -1,3 +1,7 @@
 {
-    "*": ["*"]
+    "*": ["*"],
+    "acks": ["read", "write", "*"],
```
nicole-campbell12, line 3: @dkuc spoke to Rob about this - "acks" isn't customer facing friendly / should the resource type here instead be something like "rules" where "write" permissions give user the ability to enable or disable rules and "read" lets you view rules? Also is it still "rules" or is it now "recommendations"? Recommendations is what is showing up in the cloud.redhat.com UI left nav under Advisor.

configs/permissions/insights.json
```diff
@@ -1,3 +1,7 @@
...
+    "preferences": ["read", "write", "*"],
```
nicole-campbell12, line 4: @dkuc let's remove this altogether - on the user preferences page, we allow users to set their own preferences for any application they have regular RBAC permissions to; no need for an extra preferences permission in any app.

configs/permissions/insights.json
dkuc: So we're pretty flexible.  We can change it in way you want.  Here are a few thoughts / concerns first though.

- On the engineering side we use rules/recommendations interchangeably. On the frontend it is recommendations
- Write on recommendations sounds like you can create/update recommendations

I think one point of friction is trying to force semantic meaning on permissions.  Roles are much better for that as they have a title and description.

configs/permissions/insights.json
dkuc: How will the advisor api know to accept a call to update preferences?

coderbydesign:
Hey @dkuc - based on https://github.com/RedHatInsights/rbac-config/pull/88 you'll want to rebase and apply these changes to `configs/permissions/advisor.json` instead of `configs/permissions/insights.json`.

configs/permissions/insights.json
nicole-campbell12: @dkuc how does the advisor API know when user preferences are updated for Advisor today? currently the user preferences page is checking for the user to have app permissions, not user preferences permissions. I'm not sure exactly how the interactions work app-side but I would assume that since this works today, it would continue to work that way and is not impacted by custom role app permissions. I may be missing some consideration so I'm happy to chat about this on a call if that's easier.

configs/permissions/insights.json
nicole-campbell12: @dkuc maybe let's just use the full word then?

dkuc:
@coderbydesign @nicole-campbell12  This should be good to go. Feel free to merge if you see fit.

nicole-campbell12 (approved):

coderbydesign (approved):

PR #84, opened 2020-08-26
[WIP] updated patch permissions config

Josca:


configs/permissions/patch.json
```diff
@@ -1,3 +1,3 @@
 {
-    "*": ["*"]
+    "all": ["all"]
```
nicole-campbell12, line 2: @Josca we're looking for permissions to be defined like this: 
 "*" : ["*"]
"advisories": ["read"] 
"export" : ["read"]
"packages" : ["read"]

Josca: Yes, but we don't want any particular permissions, just "allow all" or "hide all". What we should do in that case, please?

coderbydesign: The goal is to have a representation of what these "allow all" permissions represent for the `resource_type` and `verb` fields, so that roles can be created in the UI with those permissions in mind. There won't be an option for "all" in the UI, which was the reasoning behind having teams identity the granular permissions. Correct me if I'm wrong @nicole-campbell12 ?

We'd still have the `"*": ["*"]` permission for the system/default roles to use, but the goal would be to then have resources/actions on those resources check for either "*" or the granular permission.

nicole-campbell12: Keith, you're correct - @Josca the current canned role can continue to serve as the 'allow all / grant permissions now and future' and the more granular definitions should provide indication of what 'all' means at this point in time. org admins who want to permit all can assign the current Patch admin role; those who aren't comfortable can use custom roles to create permissions they are more comfortable with otherwise.

beav: agreed; basically just leave the `"*": ["*"]` since that's for legacy users.

If your app doesn't do anything involving writes, you could do something like `"patch": ["read"]` as the only additional permission.

Josca: @beav It's a bit different. Mostly Patch only allows to view info about updates, system and packages. It's true it also allows **delete** system manually from Patch. But we have instruction from product not to add any granular permissions, at least now. So can I update config to be:
~~~
"*": ["*"]
"patch":["*"]
~~~
? It should satisfy current product requirements. Thanks.

beav: @nicole-campbell12 re @josca's comment, would it make sense for patch to just stick with the existing `*:*` for now and not make updates to this file?

nicole-campbell12: just confirmed with Shane - Patch will continue to just have the existing wildcard permissions for now and won't need to onboard to custom roles at this time.

Josca:
Ok, closing that as no actions are needed currently from Patch team. We don't want any granular permissions.

PR #87, opened 2020-09-03
Cost mangement role updates

adberglund:
## Summary
This updates permissions and default roles for cost management.

### DIFF
--- a/configs/permissions/cost-management.json
+++ b/configs/permissions/cost-management.json
@@ -1,8 +1,10 @@
 {
     "aws.account": ["*"],
+    "aws.organizational_unit": ["*"],
     "azure.subscription_guid": ["*"],
     "openshift.cluster": ["*"],
-    "rate": ["read", "*"],
+    "openshift.node": ["*"],
+    "openshift.project": ["*"],
+    "cost_model": ["read", "*"],
     "*": ["*"],
-    "aws.organizational_unit": ["*"]
 }
--- a/configs/permissions/cost-management.json
+++ b/configs/permissions/cost-management.json
@@ -1,10 +1,10 @@
 {
-    "aws.account": ["*"],
-    "aws.organizational_unit": ["*"],
-    "azure.subscription_guid": ["*"],
-    "openshift.cluster": ["*"],
-    "openshift.node": ["*"],
-    "openshift.project": ["*"],
-    "cost_model": ["read", "*"],
+    "aws.account": ["*", "read"],
+    "aws.organizational_unit": ["*", "read"],
+    "azure.subscription_guid": ["*", "read"],
+    "openshift.cluster": ["*", "read"],
+    "openshift.node": ["*", "read"],
+    "openshift.project": ["*", "read"],
+    "cost_model": ["*", "read", "write"],
     "*": ["*"],
 }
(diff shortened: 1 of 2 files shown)


nicole-campbell12:
@coderbydesign how will these proposed role changes impact users who already have the cost-management:rate:* permission (since it looks like that permission is being removed from the admin role and replaced with cost-management:cost_model:*)?

coderbydesign:
@nicole-campbell12 for accounts which have not updated their default group, the default role would update with the new permission, so that shouldn't be an issue. I think there could be a problem though where an account has modified their default group, meaning we won't update it with these changes. If that permission goes away, they'd only have the `rate` permission, not the `cost_model` permission. Is that what you're referring to?

nicole-campbell12:
@adberglund is this to no longer support the cost-management:rate:* permission or are you more so trying to rename this resource type from "rate" to "cost_model"? 
Since this would result in the modification of permissions within an existing role, any user on an account with custom default access who already has the cost-management:rate:* permission will be impacted by this change since you'd be changing to allow users to users with cost-management:cost_model:* instead. 
Just wanting to double check - is this the intended impact?

adberglund:
@nicole-campbell12 @coderbydesign , this is a rename. Doing this to reflect what is actually being RBAC'd in the cost management application. If we could still do this kind of thing, I'd say it is worth a quick inspection of the existing custom roles in the database. I'm going to guess no one has a custom role with rates for cost management. 

But yes, I assumed this would be the impact.

nicole-campbell12:
@adberglund it's not straight forward to get the number of custom roles that exist with the "rate" resource type to determine impact of this change. Are you saying that you don't think there are any and therefore the impact is likely negligible? Do you plan to respect both of these permissions to avoid any potential breaking change for existing users (if there is even a possibility) with "rate" in their permission string?

configs/permissions/cost-management.json
```diff
@@ -1,8 +1,10 @@
 {
     "aws.account": ["*"],
```
nicole-campbell12, line 2: the UI is going to filter out "*" and only show granular resource types and operations, so we'll also need to see those operations defined for each of these resource types. For example, 

` "aws.account": ["*" , "read"],`

adberglund:
@nicole-campbell12 I figured that wasn't straight forward. Yes, that is what I am saying. I don't think anyone has defined a custom "rate" role to date for cost. We can absolutely support both just to be careful. I will do that in my related PR for cost. 

Fixing the other comment here now.

nicole-campbell12:
@coderbydesign Cost plans to respect both 'rate' and 'cost model' just to be on the safe side - can you please review the PR based on that plan and see if things are set up to align with that plan?

nicole-campbell12:
@adberglund things here look good - just want to make sure Cost is already checking for both cost-management:cost_model:* and cost-management:rate:* before this PR gets merged to CI. Is there a PR to reference for the additional check?

nicole-campbell12 (approved):
do not merge until we confirm cost-side API updates for checking both rate and cost-model permissions

coderbydesign:
> @coderbydesign Cost plans to respect both 'rate' and 'cost model' just to be on the safe side - can you please review the PR based on that plan and see if things are set up to align with that plan?

Based on our conversation today, yes! We'll hold until confirmation from @adberglund

coderbydesign (approved):

adberglund:
@coderbydesign @nicole-campbell12 cost PR for this https://github.com/project-koku/koku/pull/2380 . I should be able to merge the cost PR to CI tomorrow.

PR #88, opened 2020-09-09
Update Insights admin role to 'Advisor'

wcmitchell:
Updated configs for Advisor admin role to use correct application name.

Signed-off-by: Chris Mitchell <cmitchel@redhat.com>


wcmitchell:
This will need some checking to ensure it doesn't break Advisor current setup and needs to wait for the UI to list roles by display_name, which is still in progress. Adding `no merge` flag until those are in place.

configs/roles/insights.json
```diff
@@ -2,13 +2,14 @@
...
+      "display_name": "Advisor administrator",
+      "description": "Perform any available operation against any advisor resource.",
       "system": true,
       "platform_default": true,
-      "version": 6,
+      "version": 7,
       "access": [
         {
-          "permission": "insights:*:*"
+          "permission": "advisor:*:*"
```
astrozzc, line 12: Should we modify the one in the permissions folder? We might have to think of a way to delete the old permission "insights:*:*"

wcmitchell: Yep, you're right. I just realized that looking at another PR. For the short term, I'm thinking of just adding an `advisor.json` file in the permissions directory.

astrozzc (approved):
LGTM!

wcmitchell:
Blockers cleared, sending to CI

PR #93, opened 2020-09-17
Add inventory read permissions to App Admin roles

iphands:
Inventory read permissions are now a thing. Adding this permission to admin Roles
for apps that have Inventory views and call /api/inventory to draw them.

Without this, a customer would see screens like this: https://marvelapp.com/prototype/f3274de/screen/72596828

Note: we will still have scenarios where customers see ^ so we still need to do ^
Because custom roles can let you setup folks w/o those permissions.

### DIFF
--- a/configs/roles/subscriptions.json
+++ b/configs/roles/subscriptions.json
@@ -8,7 +8,8 @@
       "version": 2,
       "access": [
         {
-          "permission": "subscriptions:*:*"
+          "permission": "subscriptions:*:*",
+          "permission": "inventory:*:read"
         }
       ]
     }
--- a/configs/roles/compliance.json
+++ b/configs/roles/compliance.json
@@ -8,7 +8,8 @@
       "version": 6,
       "access": [
         {
-          "permission": "compliance:*:*"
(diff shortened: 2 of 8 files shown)


coderbydesign:
This makes sense to me. cc: @nicole-campbell12

configs/roles/subscriptions.json
```diff
@@ -8,7 +8,8 @@
       "version": 2,
       "access": [
         {
-          "permission": "subscriptions:*:*"
+          "permission": "subscriptions:*:*",
+          "permission": "inventory:*:read"
```
nicole-campbell12, line 12: @iphands just double checking on this - thought we chatted about adding read-only inventory permissions to all insights apps except for subs watch?

iphands: I'll check with them on this

iphands:
@coderbydesign hold off on merge I want to ask folks before

iphands:
This can be shipped all the way to prod! :D

coderbydesign (approved):
Approving and merging per @nicole-campbell12

PR #94, opened 2020-09-24
NOTIF-17

jeromemarc:
Added granual permissions for Notifications (requested by Nicole).

### DIFF
--- a/configs/permissions/notifications.json
+++ b/configs/permissions/notifications.json
@@ -1,5 +1,4 @@
 {
     "*": ["*"],
-    "notifications": ["read","write"],
-    "integrations": ["read","write"]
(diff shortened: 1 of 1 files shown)


nicole-campbell12:
@jeromemarc just double checking that you want to combine permissions for notifications AND integrations within the same notifications namespace. This would mean in the UI, notifications:notifications:read and notifications:integrations:read would should up (for example) if a user filters on Notifications as the app. 

If not, you can create a new PR that defines permissions for Integrations that would could include in the canned role for Notifications if you want to keep them together in that way. Up to you - just let me know.

nicole-campbell12 (approved):

coderbydesign (approved):

PR #98, opened 2020-10-01
Remove insights permission file

coderbydesign:
With the changes in https://github.com/RedHatInsights/rbac-config/pull/88/files we
migrated the insights permission and role name to `advisor` vs `insights`. This
removes the old insights permission file.


astrozzc (approved):
LGTM

PR #99, opened 2020-10-01
Move config to QA

coderbydesign:
Before merging, we need to verify which changes are moving up, and which need to stay back. We'll also hope to get #87 and #94 in.


coderbydesign:
Closing in favor of #102

PR #103, opened 2020-10-08
Fix json syntax error

Victoremepunto:


astrozzc (approved):
Nice! LGTM

configs/permissions/cost-management.json
```diff
@@ -6,5 +6,5 @@
     "openshift.node": ["*", "read"],
     "openshift.project": ["*", "read"],
     "cost_model": ["*", "read", "write"],
-    "*": ["*"],
```
astrozzc, line 9: Thanks for fixing this! This extra "," causes jason load error when seeding permissions, which is introduced by: https://github.com/RedHatInsights/rbac-config/commit/f21ddb4e2e9c5c9f384a0683b25ced07e5ed8ec5#diff-d63abce8f26658d70a68c62f2eb88368

coderbydesign (approved):
Thanks, @Victoremepunto!

PR #105, opened 2020-10-21
Verb change

astrozzc:
Update config of cost-managment permission
    
    The new cost-managment permissions are staying in CI. So
    separate the update in a new commit.

Change structure of permission config to support description
    
    Change the list of verbs from string to json so support adding description to
    the permission.
    
    JIRA: https://issues.redhat.com/browse/RHCLOUD-8979


wcmitchell (approved):
👍

configs/permissions/cost-management.json
```diff
@@ -1,10 +1,35 @@
...
+    ],
+    "openshift.project": [
+        {"verb":"*"},
+        {"verb":"read"}
+    ],
+    "cost_model": [
+        {"verb":"*"},
+        {"verb":"read"},
+        {"verb":"write"},
+        {"verb":"read"}
```
coderbydesign, line 30: Looks like a duplicate here?

astrozzc: Good catch! Updated

coderbydesign (commented):
I checked these out locally and things seed as expected. One nit might be to format the JSON files? It might help with linting diffs moving forward.

coderbydesign (approved):

PR #108, opened 2020-11-20
NOTIF-16: Add 'Integrations administrator' role & permissions

theute:


configs/permissions/integrations.json
```diff
@@ -0,0 +1,4 @@
+{
+    "*": ["*"],
+    "endpoints": ["read","write"]
```
coderbydesign, line 3: Just to clarify expectations, this new file would be creating the following permissions:
```
integrations:*:*
integrations:endpoints:read
integrations:endpoints:write
```

configs/roles/integrations.json
```diff
@@ -0,0 +1,16 @@
+{
+  "roles": [
+    {
+      "name": "Integrations administrator",
+      "description": "Perform any available operation against any Integrations resource.",
+      "system": true,
+      "platform_default": false,
```
coderbydesign, line 7: The "Notifications administrator" role is a platform default. Integrations does not need to be?

configs/roles/integrations.json
```diff
@@ -0,0 +1,16 @@
...
+      "version": 1,
```
coderbydesign, line 8: The role versions actually need to be started at `"version": 2` due to some nuance in the way our seeds run.

coderbydesign (changes requested):
Hey @theute! Thanks for the PR. I left a couple of comments for clarification, and one minor update that's needed.

configs/permissions/integrations.json
theute: Correct (we're not using those yet, but should at some point)

configs/roles/integrations.json
theute: Correct.
- Integrations are supposed to be created only by account admins (for now it's a all or nothing)
- Any user can define his/her notifications

configs/roles/integrations.json
theute: I had no clue :-)

theute:
Should be good now

coderbydesign (approved):

PR #111, opened 2021-01-11
NOTIF-102 Adapt to new requirements

theute:
In addition, we'll need:

- Remove "Notifications administrator" role from database, I changed to "Notifications Administrator" to be a bit more consistent (but not everyone follows this one)
- Remove "Integrations administrator" role from database, this no longer exists
- Maybe delete the "*" permissions from database ?

PS: I don't know what: `"system": true` is, I just used the same as for all other roles.


theute:
Needs review from:
- @jeromemarc 
- @pilhuhn

pilhuhn (approved):
LGTM

jeromemarc:
LGTM

configs/roles/notifications.json
```diff
@@ -2,15 +2,34 @@
   "roles": [
     {
       "name": "Notifications administrator",
-      "description": "Perform any available operation against any Notifications resource.",
+      "description": "Perform any available operation against notifications and integrations applications.",
```
nicole-campbell12, line 5:
```suggestion
      "description": "Perform any available operation against Notifications and Integrations applications.",
```

configs/permissions/notifications.json
```diff
@@ -1,15 +1,15 @@
...
-    ],
     "notifications": [
         {
             "verb": "read"
         },
         {
             "verb": "write"
         }
+    ],
+    "*": [
```
pilhuhn, line 10: Is it documented somewhere that in such a case the specific case has precedence over `"*"` ?

theute:
@lphiri @coderbydesign @nicole-campbell12 do we need the *.* permissions in the `permissions/*` files like they are defined here ? looks like those are removed from the list of permission one can assign to a role through the UI

(Understood that we want them in roles/notifications.json, I just don't know if they also need to be in the `permissions` files)

coderbydesign:
@theute yes, we'd still prefer the permissions files to be the source of truth for an app's permissions, even those filtered out by the UI.

wcmitchell (approved):
LGTM

PR #121, opened 2021-02-22
Setup push action for `master` to convert JSON to ConfigMaps

coderbydesign:
This will use the following action: https://github.com/RedHatInsights/rbac-config-actions
to convert over JSON config into ConfigMaps when a new commit is pushed to `master`,
and will push a commit with the file changes up based on the output of the action.

By default, these will reside in in `_private/configmaps/*` - note that we'll need to update the 
action once we namespace the config, but this will help us test the integration until then.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -1,0 +1,16 @@
+on:
+  push:
+    branches:
+      - master
+name: Config Conversion
+jobs:
+  convert_config:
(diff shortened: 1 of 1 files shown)


casey-williams-rh (approved):
LGTM

astrozzc:
LGTM

astrozzc (approved):

PR #126, opened 2021-02-24
playbook-dispatcher permissions and remediations role update

jharting:
https://issues.redhat.com/browse/RHCLOUD-11953

### DIFF
--- a/configs/ci/playbook-dispatcher.json
+++ b/configs/ci/playbook-dispatcher.json
@@ -1,0 +1,10 @@
+{
+    "run": [
+        {
+            "verb": "read"
+        },
+        {
+            "verb": "write"
+        }
+    ]
+}
--- a/configs/ci/roles/remediations.json
+++ b/configs/ci/roles/remediations.json
@@ -5,13 +5,25 @@
       "description": "Perform any available operation against any Remediations resource",
       "system": true,
       "platform_default": false,
-      "version": 4,
+      "version": 5,
       "access": [
         {
           "permission": "remediations:*:*"
         },
         {
           "permission": "inventory:*:read"
+        },
+        {
+          "permission": "playbook-dispatcher:run:read",
+          "resourceDefinitions": [
+            {
+              "attributeFilter": {
+                "key": "service",
+                "operation": "equal",
+                "value": "remediations"
+              }
+            }
+          ]
         }
       ]
     },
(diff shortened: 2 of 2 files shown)


configs/permissions/playbook-dispatcher.json
```diff
@@ -0,0 +1,10 @@
+{
+    "run": [
```
jharting, line 2: @coderbydesign is the permission definition OK like this or do I also need to define the various permutations with wildcards?

coderbydesign: You won't have any system/canned roles with any of these permissions available, this is just for custom role creation? If so, these granular permissions should be all you need! Custom role creation doesn't use wildcards. 

I'll let @nicole-campbell12 correct me if I'm wrong here. 😄

jharting: These permissions will be referenced from system/canned roles (e.g. the remediations role below, plus others) however neither of these needs wildcards.

coderbydesign:
@jharting could you rebase this on `master` when you're able? It should resolve the JSON validation check that's failing.

configs/roles/remediations.json
```diff
@@ -27,6 +39,18 @@
         },
         {
           "permission": "remediations:remediation:write"
+        },
+        {
+          "permission": "playbook-dispatcher:run:read",
```
coderbydesign, line 44: @jharting it looks like this is identical to the entry above, is this supposed to be `"permission": "playbook-dispatcher:run:write",` here?

jharting: @coderbydesign 
No, it's on purpose. I was trying to make the roles self-contained i.e. if I assign a user to `Remediations administrator` they should get full access even without having `Remediations user` (i.e. it should not be assumed that the administrator already has read permission from the Remediations user role). LMK if that's wrong.

coderbydesign: No that makes sense. I was reading this permission initially as being a part of the `Remediations administrator` role because of the way GH had the diff. 😄 

You will however want to bump the version of this role in order to pick the changes up.

jharting: @coderbydesign good catch. It should be fixed now.

coderbydesign (approved):

PR #156, opened 2021-03-30
Condense catalog and approval administrator roles using wildcards

eclarizio:
@bzwei @hsong-rh @mkanoor Please review and make sure this is ok.

/cc @hyperkid123 This should help keep the permissions in sync.


configs/ci/roles/approval.json
```diff
@@ -7,141 +7,7 @@
...
-                "key": "scope",
-                "operation": "equal",
-                "value": "group"
-              }
-            }
-          ]
-        },
-        {
-          "permission": "approval:actions:read",
+          "permission": "approval:*:*",
```
mkanoor, line 10: @eclarizio By doing a wildcard we loose the ability to control which are the objects under rbac, and how they should be handled differently for regular user versus the admin.

bzwei: Although I cannot recall the details, we did not use wildcards for good reasons.

mkanoor: Wouldn't that defeat the whole purpose of it.
What is the issue that we are trying to solve.

 Is it to support any new objects that get added to catalog/approval are not listed in this config file. 
Or is it to support non RBAC objects?

Any object that gets added to the system needs to be reviewed for both Admin and regular user access and set appropriately. If we just keep using * for admin we have never discussed what happens to the user.

Hyperkid123: The catalog has a requirement for the landing page content to gate some of it behind catalog/approval admin persona. But officially there are no personas. There is no catalog admin flag in the user object or something similar. The only way to determine the persona right now is by permissions. Catalog admin has I think 28 unique permissions and we have to keep that list somewhere.

Let's say you add a new permission to catalog admin. At this point, the list we have stored on the landing page, is not complete, which will mean the content won't be displayed even though it should be. So we will constantly have to keep this in sync.

mkanoor: When the user logs in, can the UI make a call to get the current Access for the user using this RBAC API call.
https://github.com/RedHatInsights/insights-rbac/blob/cd528d4199435008f1678b1ea8bbca34bacb1cc4/docs/source/specs/openapi.json#L2315

If this list which is coming from the server is being compared with some thing that is statically stored in the app, shouldn't that data be externalized.

I think the RBAC team at one point was trying to figure out if the objects and permissions for an app can be exposed.

mkanoor: I am not sure if this is the file https://github.com/RedHatInsights/rbac-config/blob/3711aebb2ad227210f4492e655dff54bdc15f596/configs/prod/permissions/catalog.json#L1

Hyperkid123: Yeah, we actually have all user permissions available in our user object. The part I am "worried" about is the static list we have to compare it to.

If we could get the specific catalog/approval admin role permissions that would be great. I think it could be done via https://github.com/RedHatInsights/insights-rbac/blob/cd528d4199435008f1678b1ea8bbca34bacb1cc4/docs/source/specs/openapi.json#L1964 but before we can get that we would need to make at least additional 2 sequential API calls which would slow the content loading even further.

mkanoor: Have you tried. this API end point
https://ci.cloud.redhat.com/api/rbac/v1/permissions/?application=catalog

Hyperkid123: Well yes but that does not solve the problem of personas. It gives the list of catalog permissions, but not the list of permissions for catalog/approval admin/user.

coderbydesign: Just hopping into this conversation without context, but if you're looking for the permissions for a role, you can retrieve them via the `access` object in a role detail call such as: https://cloud.redhat.com/api/rbac/v1/roles/_UUID_/

coderbydesign:
Can this be closed out?

mkanoor:
@coderbydesign I think so. We can reopen if need be.

coderbydesign:
Great, thanks @mkanoor!

PR #130, opened 2021-03-08
Add IBM account for Cost RBAC (CI/QA only)

boaz0:
Don't merge before https://github.com/project-koku/koku/pull/2710

//cc @adberglund @dccurtis

PR #157, opened 2021-03-31
Bump Cost Cloud Viewer role in stage/prod

coderbydesign:
PR https://github.com/RedHatInsights/rbac-config/pull/112 added `gcp` permissions
to the Cost Cloud Viewer role, but didn't bump the version, so the changes were not
applied in stage or production.

They are fine in CI and QA because of https://github.com/RedHatInsights/rbac-config/pull/130
which added the IBM permission to the roles in those environments and bumped the version.


wcmitchell (approved):
👍

PR #162, opened 2021-04-15
RHC roles, config-manager permissions

Jason-RH:


configs/ci/roles/config-manager.json
```diff
@@ -0,0 +1,67 @@
...
+          "permission": "inventory:*:read"
+        },
+        {
+          "permission": "playbook-dispatcher:run:read",
+          "resourceDefinitions": [
+            {
+              "attributeFilter": {
+                "key": "service",
+                "operation": "equal",
+                "value": "config-manager"
```
jharting, line 29: @Jason-RH I had configured the service predicate value as `config_manager` in playbook-dispatcher. I can change that there or, if you don't mind, change it here so that we get a match.

Jason-RH: :+1: I can change it here

jharting (approved):

Jason-RH:
@coderbydesign I went ahead and removed the configs for stage/prod so that we can get these added in stages.

coderbydesign (approved):

PR #164, opened 2021-04-22
Add fine-grained resources for Subscriptions app

patrickeasters:
Adds these new permissions in ci/qa/stage:
- Read Reports
- Read/Write Manifests

Makes these role changes:
- Renames "Subscription Watch administrator" to "Subscriptions administrator" with read/write permissions and makes it non-default
- Adds a "Subscription Watch user" role with view-only permissions

### DIFF
--- a/configs/ci/roles/subscriptions.json
+++ b/configs/ci/roles/subscriptions.json
@@ -2,13 +2,32 @@
   "roles": [
     {
       "name": "Subscription Watch administrator",
-      "description": "Perform any available operation against any Subscription Watch resource.",
+      "display_name": "Subscriptions administrator",
+      "description": "Perform any available operation against any Subscriptions resource.",
       "system": true,
-      "platform_default": true,
-      "version": 4,
+      "platform_default": false,
+      "version": 5,
       "access": [
         {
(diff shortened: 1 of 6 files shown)


configs/ci/roles/subscriptions.json
```diff
@@ -2,10 +2,11 @@
...
+      "display_name": "Subscriptions administrator",
+      "description": "Perform any available operation against any Subscriptions resource.",
       "system": true,
-      "platform_default": true,
-      "version": 4,
+      "platform_default": false,
+      "version": 5,
       "access": [
         {
           "permission": "subscriptions:*:*"
```
coderbydesign, line 12: Hey, @patrickeasters - based on the email thread with Nicole, is the admin role meant to have `subscriptions:manifests:write` now as well?

patrickeasters: Yea, it's intended to have that now via the implicit `subscriptions:*:*`

coderbydesign: 👍

coderbydesign (approved):

PR #166, opened 2021-04-28
Config manager stage prod

Jason-RH:
@coderbydesign These roles are currently working in CI/QA and I'd like to get them added to Stage/Prod. Is it okay to keep them as version 2 or should I update all config-manager roles to version 3 (no changes, just new environments)?


coderbydesign:
@Jason-RH that should be fine!

coderbydesign (approved):

PR #174, opened 2021-06-09
Add ROS app configs in CI

apuntamb:
Introduce rbac configs for ROS app. Adding read only permissions currently.
Adding @beav into the loop for confirmation about the permissions/roles required for ROS.

Edit: Adding @kgaikwad in the loop for review by teammate.


configs/ci/roles/ros.json
```diff
@@ -0,0 +1,19 @@
+{
+    "roles": [
+        {
+          "name": "ROS Administrator",
+          "description": "Perform any available operation against any Resource Optimization service.",
+          "system": true,
+          "platform_default": true,
+          "version": 8,
```
coderbydesign, line 8: It doesn't really matter, but just curious if this is version 8 for a reason?

configs/ci/roles/ros.json
```diff
@@ -0,0 +1,19 @@
+{
+    "roles": [
+        {
+          "name": "ROS Administrator",
+          "description": "Perform any available operation against any Resource Optimization service.",
+          "system": true,
+          "platform_default": true,
```
coderbydesign, line 7: Is the expected behavior to give everyone admin access by default?

configs/ci/permissions/ros.json
```diff
@@ -0,0 +1,12 @@
+{
+    "*": [
+        {
+            "verb": "*"
+        }
+    ],
+    "system": [
```
coderbydesign, line 7: Just to confirm, you'd want to add a granular permission which would be available during custom role creation, for: `ros:system:read`?

configs/ci/roles/ros.json
```diff
@@ -0,0 +1,19 @@
+{
+    "roles": [
+        {
+          "name": "ROS Administrator",
```
coderbydesign, line 4: For consistency on the platform, we'd ask that this be a lowercase `a` for `administrator`.
```suggestion
          "name": "ROS administrator",
```

configs/ci/roles/ros.json
apuntamb: We want the app to provide Read access to every user (may be accept the administrator).

configs/ci/roles/ros.json
apuntamb: sure

configs/ci/roles/ros.json
apuntamb: It should be 1, right? Since the first version of ROS  rbac config.

configs/ci/permissions/ros.json
apuntamb: We want to add read permission to all the users throughout. Do we need a custom role in this case?

configs/ci/roles/ros.json
kgaikwad: @coderbydesign, 
Here, I would like to understand meaning of version.
Is it something similar to how database schema version works? I mean in this case for particular role object. Each time, one needs to update version value to reflect new changes to particular role object, right?

configs/ci/roles/ros.json
kgaikwad: @coderbydesign, 
As @apuntamb mentioned in case of ROS, we want to provide read access to every user. Mainly ROS backend is mainly r/o.

Any suggestions on defining a role? 
For now,  can we just go with admin role or do we need to define two roles i.e admin & user where user role will be added by default to each user?

configs/ci/roles/ros.json
coderbydesign: Since these config files are used to seed data for all tenants, whenever there's a change made to the config, you also need to bump the version number in order for our seeding logic to know there is a change which should be applied.

configs/ci/roles/ros.json
coderbydesign: Ok, yes you'd need a role such as "Resource Optimization user", which would have the `ros:*:read` permission, and `platform_default: true`, vs having this administrator role be a platform defaultt.

configs/ci/permissions/ros.json
coderbydesign: The permissions files are just defining your granular permissions which can be used in the UI during custom role creation. So in this case `ros:system:read` would be the only custom role that could be created in the UI. If you wanted `ros:*:read` as well, you'd need to add `"verb": "read"` before/after line 4.

configs/ci/roles/ros.json
kgaikwad: Thank you @coderbydesign for details. 
 
@apuntamb - could you please set version value to 1?

configs/ci/permissions/ros.json
```diff
@@ -0,0 +1,10 @@
+{
+    "*": [
+        {
+            "verb": "*"
+        },
+        {
+            "verb": "read"
+        }
```
apuntamb, line 8: @coderbydesign like this you meant?

apuntamb:
@coderbydesign Could you give it another round of read?

configs/ci/permissions/ros.json
coderbydesign: That appears to align with the permissions you have in the roles, so yes! Just a note, custom role creation doesn't bring in any wildcard permissions, so you'd not be able to create any custom roles in the UI based on these permissions, you'd just have the system roles in your /roles/ros.json file, which I assume is what you want?

coderbydesign:
cc: @nicole-campbell12

configs/ci/roles/ros.json
```diff
@@ -0,0 +1,35 @@
...
+            "permission": "ros:*:*"
+          },
+          {
+            "permission": "inventory:*:read"
+          }
+        ]
+      },
+      {
+        "name": "Resource Optimization User",
+        "description": "View any ROS resource.",
```
coderbydesign, line 21: Could you also expand `ROS` in the descriptions?

configs/ci/roles/ros.json
```diff
@@ -0,0 +1,35 @@
+{
+    "roles": [
+      {
+        "name": "Resource Optimization Administrator",
+        "display_name": "Resource Optimization administrator",
+        "description": "Perform any available operation against any ROS resource.",
```
coderbydesign, line 6: Could you also expand `ROS` in the descriptions?

coderbydesign (changes requested):

configs/ci/permissions/ros.json
apuntamb: I would like to confirm this with @kgaikwad once. Are we okay with this?

kgaikwad: @coderbydesign - yes, that's fine! Currently we just need system roles that are defined in /roles/ros.json file.

configs/ci/roles/ros.json
```diff
@@ -0,0 +1,35 @@
+{
+    "roles": [
+      {
+        "name": "Resource Optimization Administrator",
```
coderbydesign, line 4: For consistency across other admin roles, this should be lowercase.
```suggestion
        "name": "Resource Optimization administrator",
```

configs/ci/roles/ros.json
```diff
@@ -0,0 +1,35 @@
...
+        "display_name": "Resource Optimization administrator",
```
coderbydesign, line 5: We don't need `display_name` here. This is just a mechanism for overriding the default name if necessary, otherwise RBAC will automatically set `display_name` to `name`.

coderbydesign (changes requested):

coderbydesign (approved):

PR #179, opened 2021-06-16
[DRFT-326] Add notify permission to drift

CodeHeeler:
Add permission for users to add/remove associations between baselines and systems for notifications purposes

### DIFF
--- a/configs/ci/permissions/drift.json
+++ b/configs/ci/permissions/drift.json
@@ -21,5 +21,10 @@
         {
             "verb": "read"
         }
+    ],
+    "notifications": [
+	{
+	    "verb": "write"
+	}
     ]
 }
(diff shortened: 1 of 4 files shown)


petrblaho:
@CodeHeeler it looks like we can't define our own verb here...

CodeHeeler:
> @CodeHeeler it looks like we can't define our own verb here...

@coderbydesign is this the case?  Is there another way we should be approaching this?  We have been having these per service, but now we need another permission within the baseline service so I thought I could just add it this way.  Does it need to be at the higher level like the app permissions and just "read/write" only?

CodeHeeler:
After Mike shared examples from other teams combined with Petr's thoughts, went ahead and updated this without waiting on Keith's feedback.

petrblaho (approved):

CodeHeeler:
Well, looks like your review wasn't enough, so requesting an approving review in addition.

tahmidefaz (approved):
:mage_man:

CodeHeeler:
> mage_man

lol, thanks!

PR #189, opened 2021-07-07
Add OCP Advisor roles and permissions

gkarat:
Fixes https://issues.redhat.com/browse/CCXDEV-5246

There is a new application under c.r.c. called OCP Advisor, which needs RBAC config set up.

### DIFF
--- a/configs/ci/permissions/ocp-advisor.json
+++ b/configs/ci/permissions/ocp-advisor.json
@@ -1,0 +1,22 @@
+{
+  "*": [
+    {
+      "verb": "*"
+    }
+  ],
+  "toggle-recommendations": [
+    {
+      "verb": "write"
+    }
+  ],
+  "recommendation-results": [
+    {
+      "verb": "read"
+    }
+  ],
+  "exports": [
+    {
+      "verb": "read"
+    }
+  ]
+}
--- a/configs/ci/roles/insights.json
+++ b/configs/ci/roles/insights.json
@@ -15,19 +15,6 @@
           "permission": "inventory:*:read"
         }
       ]
-    },
-    {
-      "name": "OCP Advisor administrator",
-      "display_name": "OCP Advisor administrator",
-      "description": "Perform any available operation against any OCP Advisor resource.",
-      "system": true,
-      "platform_default": true,
-      "version": 1,
-      "access": [
-        {
-          "permission": "ocp-advisor:*:*"
-        }
-      ]
     }
   ]
 }
--- a/configs/ci/roles/insights.json
+++ b/configs/ci/roles/insights.json
@@ -2,11 +2,11 @@
   "roles": [
     {
       "name": "Insights administrator",
-      "display_name": "Advisor administrator",
-      "description": "Perform any available operation against any advisor resource.",
+      "display_name": "RHEL Advisor administrator",
+      "description": "Perform any available operation against any RHEL Advisor resource.",
       "system": true,
       "platform_default": true,
-      "version": 9,
+      "version": 10,
       "access": [
         {
           "permission": "advisor:*:*"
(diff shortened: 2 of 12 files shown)


gkarat:
@AllenBW, could we change the `display_name` of RHEL Advisor administrator from "Advisor administrator" to "RHEL Advisor administrator"? Plus, it will similarly apply to the `description` field.

Furthermore, we could rename the RHEL Advisor application name (in terms of RBAC) from "advisor" to the "rhel-advisor" so that we distinguish them better.
  
https://github.com/RedHatInsights/rbac-config/blob/316a9cd183977ccb80398d6e375aa1f0c3546ddb/configs/prod/roles/insights.json#L5
https://github.com/RedHatInsights/rbac-config/blob/316a9cd183977ccb80398d6e375aa1f0c3546ddb/configs/prod/roles/insights.json#L6
https://github.com/RedHatInsights/rbac-config/blob/316a9cd183977ccb80398d6e375aa1f0c3546ddb/configs/prod/roles/insights.json#L12

AllenBW:
@gkarat not my call, not sure who maybe @dkuc  knows?

dkuc:
Changing description is probably fin.


Changing the display name - this might be okay.  I'm not sure how rbac stores its data.  If display name is the primary key, this might wipe away all the rbac settings users have already set up.  @coderbydesign Do you know if its safe to switch display_name?


Changing the permission -  I'd rather avoid this if we can - would require code changes throughout advisor.  I don't think this is customer visible so probably not worth the effort.

gkarat:
@dkuc, from the README, it seems that `display_name` is not a primary key but rather utilized only by UI (https://github.com/RedHatInsights/rbac-config#canned-roles): `If you need to update the name of a role, add or update the display_name field in the role config, as this is the the database value displayed in the UI.`

gkarat:
@coderbydesign, @dkuc, PTAL 🙏🏼

coderbydesign:
@gkarat @dkuc - That's correct, the `name` field is the primary key here, and `display_name` is safe to be updated. cc: @nicole-campbell12

coderbydesign (changes requested):
Hey @gkarat!

> Changing the permission - I'd rather avoid this if we can - would require code changes throughout advisor. I don't think this is customer visible so probably not worth the effort.

I agree with @dkuc here. It would require data migrations on the RBAC side as well. 

I'd also recommend moving the role into its own namespaced file, for `/roles/ocp-advisor.json` to keep them separate from the advisor roles, similar to what you did with the permissions.

gkarat:
@coderbydesign, thanks for your suggestion. PTAL at my last commit :-)

dkuc (approved):

gkarat:
@coderbydesign, is something missing/blocking us? If no, could we merge it then? 😇

coderbydesign:
@gkarat - Generally we'd have a PR only for the environment(s) you are ready to move these to, since we have a process that syncs the generated ConfigMaps into each environment updated. I'd recommend only updating whichever pre-production environment(s) you're ready to test first, and then a follow-up for stage and/or prod. 

I've added you as a watcher on the [infra ticket](https://issues.redhat.com/browse/RHCLOUD-15419) as well, for when we do promote this. Let me know if you have any questions!

gkarat:
> @gkarat - Generally we'd have a PR only for the environment(s) you are ready to move these to, since we have a process that syncs the generated ConfigMaps into each environment updated. I'd recommend only updating whichever pre-production environment(s) you're ready to test first, and then a follow-up for stage and/or prod.
> 
> I've added you as a watcher on the [infra ticket](https://issues.redhat.com/browse/RHCLOUD-15419) as well, for when we do promote this. Let me know if you have any questions!

yeah, thank you! I've created the three separate PR, PTAL https://github.com/RedHatInsights/rbac-config/pulls/gkarat

closing this one 😄

PR #197, opened 2021-07-21
Add OCP Advisor roles and perms for ci env

gkarat:
Fixes https://issues.redhat.com/browse/CCXDEV-5246

There is a new application under c.r.c. called OCP Advisor, which needs RBAC config set up.


gkarat:
@coderbydesign , PTAL 👋🏼

coderbydesign (approved):

PR #202, opened 2021-08-04
Add Malware Detection application

dkuc:
Rbac config for the new insights application Malware Detection.  For version one there will be two roles/permissions 
* An admin role
* A read only role


coderbydesign (changes requested):
@dkuc - To confirm, you're wanting the admin role/perm `malware-detection:*:*` to be given to everyone by default (this will _not_ be given to users who have already modified their default access group), with the ability to use `malware-detection:disable-signature:write` as a granular permission in custom role creation?

One thing worth updating would the the file name for the permission files. We should have it `malware-detection` since the permissions are dynamically generated based off the file name. Otherwise it would be `malware_detection:disasble-signature:write`.

coderbydesign:
cc: @nicole-campbell12

dkuc:
> One thing worth updating would the the file name for the permission files.

Done.

dkuc:
Removed all platform default roles.

dkuc:
Consolidated to an admin and reader role.

dkuc:
Can you give a another review @coderbydesign ?

dkuc:
@coderbydesign @nicole-campbell12  When do you think we can merge this in?

coderbydesign (approved):

PR #205, opened 2021-09-07
Add organization resource for subscriptions app

bharat-rajani:
#### Adds these new permissions in ci/qa/stage:

- Read/Manage Organization

#### Makes these role changes:

- Modifying "Subscriptions user" role to have `read` access to organization resource

### DIFF
--- a/configs/ci/permissions/subscriptions.json
+++ b/configs/ci/permissions/subscriptions.json
@@ -12,6 +12,14 @@
             "verb": "write"
         }
     ],
+    "organization": [
+        {
+            "verb": "read"
+        },
+        {
+            "verb": "manage"
+        }
+    ],
     "*": [
         {
             "verb": "*"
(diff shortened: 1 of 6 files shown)


configs/ci/permissions/subscriptions.json
```diff
@@ -12,6 +12,14 @@
             "verb": "write"
         }
     ],
+    "organization": [
+        {
+            "verb": "read"
+        },
+        {
+            "verb": "manage"
```
patrickeasters, line 20: After meeting with the platform team, we have decided to keep using the `write` permission to be consistent with the rest of the platform.

They are planning to do some work in the future to allow configuring implied permissions, which should help act as a guard rail for users creating their own custom roles.

So, if we just `s/manage/write` for these new permissions, we should be in good shape.

patrickeasters (approved):
LGTM after changing that back to `write`.

coderbydesign (approved):

PR #215, opened 2021-10-29
Prod permission and roles for subscriptions app

bharat-rajani:
#### Adds these permissions in prod (replicates lower env permissions):

- Read and Write Organization
- Read and Write Manifest
- Read Reports

#### Adds these role changes in prod (replicates lower env roles):

- **Subscription Watch administrator** role 
- **Subscriptions user** role


astrozzc (approved):
LGTM

PR #217, opened 2021-11-02
NOTIF-361 Protect Event log with new RBAC permission

bennetelli:


theute:
Looks good to me

configs/prod/roles/notifications.json
```diff
@@ -10,6 +10,9 @@
         {
           "permission": "notifications:*:*"
         },
+        {
+          "permission": "events:*:*"
```
josejulio, line 14: same as above

configs/qa/roles/notifications.json
```diff
@@ -10,6 +10,9 @@
         {
           "permission": "notifications:*:*"
         },
+        {
+          "permission": "events:*:*"
```
josejulio, line 14: and the same

configs/stage/roles/notifications.json
```diff
@@ -10,6 +10,9 @@
         {
           "permission": "notifications:*:*"
         },
+        {
+          "permission": "events:*:*"
```
josejulio, line 14: same

configs/ci/roles/notifications.json
```diff
@@ -10,6 +10,9 @@
         {
           "permission": "notifications:*:*"
         },
+        {
+          "permission": "events:*:*"
```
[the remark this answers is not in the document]
josejulio: Just to clarify, we want `events` to be part of notifications, not an app by itself.

configs/stage/roles/notifications.json
bennetelli: not sure. That was my question as well. Maybe @theute knows if this is needed or not?

josejulio: I don't think we need it

theute: @josejulio is right, format is application:<resource type>:<verb>
https://consoledot.pages.redhat.com/docs/dev/services/rbac.html

astrozzc (approved):

josejulio:
We will need to revert some changes in this PR

bennetelli:
> We will need to revert some changes in this PR

should it be notifications:events:* ? or how should it look like?

theute:
There is already `notifications:*:*` so `notifications:events:*` is already covered.
But  "events:*:*" needs to go away in roles since `events` is not an application

astrozzc:
Oops, the page is not refreshed when I merge it. Just saw the following comments. Will wait for another PR to fix it. @bennetelli Please ping me when you have the fix PR ready

bennetelli:
@astrozzc i will

PR #226, opened 2021-11-29
Cloud to console for rbac role description

john-dupuy:
In the UI, I noticed that the role description used `cloud.redhat.com` which may be confusing for users. 

cc @coderbydesign @wcmitchell @astrozzc

### DIFF
--- a/configs/ci/roles/rbac.json
+++ b/configs/ci/roles/rbac.json
@@ -3,7 +3,7 @@
     {
       "name": "User Access administrator",
       "display_name": "User Access administrator",
-      "description": "Grants a non-org admin full access to configure and manage user access to services hosted on cloud.redhat.com. This role can only be viewed and assigned by Organization Administrators.",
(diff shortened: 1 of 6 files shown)


astrozzc (approved):
Thanks for updating this, it has been here for a long time

coderbydesign (changes requested):
Thanks @john-dupuy! Could you update the PR to only modify the `configs/*` files, and not the `_private/*` files? We have a GitHub action that will update those files by generating new configMaps once the JSON is merged.

john-dupuy:
> Thanks @john-dupuy! Could you update the PR to only modify the `configs/*` files, and not the `_private/*` files? We have a GitHub action that will update those files by generating new configMaps once the JSON is merged.

Ah thanks for the review - I've just done this!

coderbydesign:
@john-dupuy sorry, was just going to merge this and realized we need to bump the versions on the roles as well.

coderbydesign (approved):
Thanks, John!

PR #229, opened 2021-12-27
RBAC permissions for activation_keys.

skuchcha:


coderbydesign:
@skuchcha to test these in pre-production environments before the changes move to prod, please break this into two PRs, one for pre-prod environment, and once validated, one for prod. Let us know if you have any questions!

skuchcha:
> @skuchcha to test these in pre-production environments before the changes move to prod, please break this into two PRs, one for pre-prod environment, and once validated, one for prod. Let us know if you have any questions!

Okay. I'll remove the changes done for prod in this PR.

configs/ci/roles/config-manager.json
```diff
@@ -8,6 +8,9 @@
       "platform_default": false,
       "version": 3,
```
coderbydesign, line 9: You'll need to bump the `"version"`  to `4` in all environment files for the seed changes to be picked up.

configs/ci/permissions/config-manager.json
```diff
@@ -1,4 +1,12 @@
 {
+    "activation_keys": [
+        {
+            "verb": "read"
+        },
+        {
+            "verb": "write"
+        }
```
coderbydesign, line 8: Even though it's not able to be used for custom role creation, we should define the global `config-manager:activation_keys:*` permission in the permissions files since they're the source of truth for available permissions, given it's being used in a role.
```suggestion
        },
        {
            "verb": "*"
        }
```

configs/ci/roles/config-manager.json
```diff
@@ -42,6 +45,9 @@
       "platform_default": true,
       "version": 3,
       "access": [
+        {
```
coderbydesign, line 48: This role in all envs will also need a version bump.

coderbydesign (approved):
Thanks for the quick responses on feedback, @skuchcha!

PR #243, opened 2022-02-17
NOTIF - remove read permissions from regular users

aferd:


josejulio:
@astrozzc could you take a look? 
We want to remove the `Notifications viewer` role from the default permissions.

configs/prod/roles/notifications.json
```diff
@@ -19,7 +19,7 @@
       "name": "Notifications viewer",
       "description": "Read only access to notifications and integrations applications.",
       "system": true,
-      "platform_default": true,
+      "platform_default": false,
       "version": 1,
```
astrozzc, line 23: Sure, you might also need to update the version of the roles in order to update the role.

BTW, current seeding process is very slow, so we paused any new seedings. We are in the process of migrating to serve from public schema, which will solve this problem. Hopefully which will be done before next week. @coderbydesign @wcmitchell

josejulio: IIUC, we might need to bump the version to `2` in both files.

@aferd  could you update?

> BTW, current seeding process is very slow, so we paused any new seedings.

Understood, thanks!

josejulio (approved):

astrozzc (approved):
LGTM!

josejulio:
@astrozzc is OK to merge now and the seeder will pick it up when it's ready?

astrozzc:
We can merge it but the seed might have to wait, we are stop using tenant schema and that should be done by next week. After that, we could start seeding, which takes about seconds

josejulio:
> We can merge it but the seed might have to wait, we are stop using tenant schema and that should be done by next week. After that, we could start seeding, which takes about seconds

that's OK, thanks

coderbydesign (approved):

theute:
For the records https://issues.redhat.com/browse/NOTIF-475

PR #244, opened 2022-02-24
Admin default schema

mjholder:
I added a new flag, admin_default, to the role schema which will be "false" by default. This will be responsible for giving role's admin permissions and allowing the PR check to validate against the schema. I also updated the README to have an example role with this new flag and a brief description of what it does.


README.md
```diff
@@ -106,6 +106,30 @@ exist for the current app/resource type.
 
 Please check existing files for more samples.
 
+Admin Role
```
astrozzc, line 109: Maybe let's call it Admin default role, to be more specific.

astrozzc (approved):
LGTM, let's see if Keith and Chris have some comments on the description of README

coderbydesign (approved):
Apart from Jay's suggestion, this looks good to me!

wcmitchell (approved):
LGTM!

PR #246, opened 2022-03-01
Add OCP Advisor roles and permissions for prod

gkarat:
Implements https://issues.redhat.com/browse/CCXDEV-7349.

There is a new application under c.r.c. called OCP Advisor which has a GA this week. This patch is another preparation step that updates the RBAC config in prod.

### DIFF
--- a/configs/prod/permissions/ocp-advisor.json
+++ b/configs/prod/permissions/ocp-advisor.json
@@ -1,0 +1,22 @@
+{
+  "*": [
+    {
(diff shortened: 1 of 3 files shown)


gkarat:
This is identical to the previous PR like for example https://github.com/RedHatInsights/rbac-config/pull/199

gkarat:
@coderbydesign , PTAL 👋🏼

coderbydesign (approved):

PR #250, opened 2022-03-28
Add display_name property for catalog roles

eclarizio:
This is meant to help fix https://issues.redhat.com/browse/SSP-2423 (which is related to https://bugzilla.redhat.com/show_bug.cgi?id=1816152)

PR #252, opened 2022-04-04
Bump catalog role versions

coderbydesign:
We need to bump the versions to pick the changes up from: https://github.com/RedHatInsights/rbac-config/pull/250/files

Disregarding ci/qa as they're no longer used.


astrozzc (approved):
LGTM

PR #265, opened 2022-05-04
Add admin_default to all HBI Admin roles

kruai:
Adds the new "admin_default" field to HBI's administrator roles in all environments.

### DIFF
--- a/configs/ci/roles/inventory.json
+++ b/configs/ci/roles/inventory.json
@@ -5,7 +5,8 @@
       "description": "Perform any available operation against any Inventory resource.",
       "system": true,
       "platform_default": true,
-      "version": 2,
+      "admin_default": true,
+      "version": 3,
       "access": [
         {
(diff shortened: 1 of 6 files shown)


coderbydesign:
@kruai - Just wanted to note that any role marked `"platform_default": true` will be given to all users by default, including org admins. Adding `"admin_default": true` to a role which is already marked `"platform_default": true` is superfluous and won't have any additional impact, since it's already inherited by all users.

If this role is intended to only be given to admins by default, then you could leave `"admin_default": true` and remove `"platform_default": true`, which would _not_ give this role to non-admins. In that case, you may want to tag another role with `"plat_default": true` if you wanted to provide a different level of access for all non-admins.

Let me know if you have any questions!

kruai:
@coderbydesign We were doing this to cover a potential future edge case, but it turns out we don't need it. Closing now!

PR #274, opened 2022-05-13
Add description of external role

astrozzc:


README.md
```diff
@@ -130,6 +130,27 @@ We added support for the new role flag "admin_default", similar to “platform_d
 }
 ```
 
+External Role
+---------------------
+We added support to define external role. In the `roles` directory, the external role could be defined as a regular role but without access/permission. The external role should have a field `external` defined, which has external id and external tenant. An example ocm role is displayed below.
```
coderbydesign, line 135:
```suggestion
We added support to define external roles. In the `roles` directory, an external role can be defined as a regular role but without access/permissions. An external should have an object `external` defined, which has external `id` and external `tenant` fields. An example ocm role is displayed below.
```

README.md
```diff
@@ -130,6 +130,27 @@ We added support for the new role flag "admin_default", similar to “platform_d
 }
 ```
 
+External Role
```
coderbydesign, line 133:
```suggestion
External Roles
```

README.md
```diff
@@ -130,6 +130,27 @@ We added support for the new role flag "admin_default", similar to “platform_d
...
+      "version": 2,
+      "external": {
+        "id": "ClusterOwner",
+        "tenant": "ocm"
+      }
+    }
+  ]
+}
+```
+After the role is seeded into system. The role could be assigned to customer through group and the external team could use the external id of the rbac role to identify it as one of the role on their side.
```
coderbydesign, line 152:
```suggestion
After the role is seeded into system, the role can be assigned to users through groups, and the external tenant team can use the external id of the RBAC role to identify it and map it to one of their roles.
```

coderbydesign (commented):
Hey @astrozzc! Just had a few wording suggestions, but was also wondering if we could add something to our [role schema](https://github.com/RedHatInsights/rbac-config/blob/master/schemas/roles.schema) to validate the fields on `external` when it's present?

coderbydesign (approved):

astrozzc:
Thanks for the suggestions!

PR #303, opened 2022-07-05
Seed OCM roles in stage

paradigm:
Required for ConsoleDot RBAC - OCM Integration

### DIFF
--- a/configs/stage/roles/ocm.json
+++ b/configs/stage/roles/ocm.json
@@ -1,0 +1,52 @@
+{
+  "roles": [
+    {
+      "name": "OCM Cluster Editor",
+      "display_name": "OCM Cluster Editor",
+      "description": "Grants permission to edit clusters",
+      "system": true,
+      "platform_default": false,
+      "version": 2,
+      "external": {
+        "id": "ClusterEditor",
+        "tenant": "ocm"
+      }
+    },
+    {
+      "name": "OCM Cluster Provisioner",
+      "display_name": "OCM Cluster Provisioner",
+      "description": "Grants permission to provision clusters",
+      "system": true,
+      "platform_default": false,
+      "version": 2,
+      "external": {
+        "id": "ClusterProvisioner",
+        "tenant": "ocm"
+      }
+    },
+    {
+      "name": "OCM Cluster Viewer",
+      "display_name": "OCM Cluster Viewer",
+      "description": "Grants permission to view clusters",
+      "system": true,
+      "platform_default": false,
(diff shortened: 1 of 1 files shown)


configs/stage/roles/ocm.json
```diff
@@ -0,0 +1,52 @@
+{
+  "roles": [
+    {
+      "name": "OCM Cluster Editor",
+      "display_name": "Cluster Editor",
+      "description": "Grants permission to edit cluster",
```
coderbydesign, line 6: Should `cluster` be plural here like the others?
```suggestion
      "description": "Grants permission to edit clusters",
```

configs/stage/roles/ocm.json
```diff
@@ -0,0 +1,52 @@
+{
+  "roles": [
+    {
+      "name": "OCM Cluster Editor",
+      "display_name": "Cluster Editor",
```
coderbydesign, line 5: Just to confirm, the `display_name` value is what will render in the UI, not the `name`. Generally these are kept the same unless the `display_name` needs to change in the UI, as `name` is the identifier for querying existing records for updates, so we're unable to modify that.

configs/stage/roles/ocm.json
```diff
@@ -0,0 +1,52 @@
...
+      "platform_default": false,
+      "version": 2,
+      "external": {
+        "id": "ClusterViewer",
+        "tenant": "ocm"
+      }
+    },
+    {
+      "name": "OCM Organization Admin",
+      "display_name": "Organization Admin",
```
coderbydesign, line 41: Similar to above, it might be a bit confusing if OCM is left off the `display_name` and even `description` fields here, since these roles will be available in the ConsoleDot UI. It may mislead them to thinking they'll have org admin permissions over ConsoleDot resources?

wcmitchell (commented):
Looks good other than Keith's CV comments

configs/stage/roles/ocm.json
paradigm: It wasn't clear to me from the user's perspective whether or not OCM is something they care about, or if it's just an implementation detail we want to obscure.  Thinking it over, it's probably better to go with including "OCM" for now; if it causes confusion, it will be more easily remedied than if it's absence caused confusion.

paradigm:
Applied suggested changes.

Once all the pieces are in place in stage, we can review the final UX/user workflow to confirm the choices here make sense before touching prod.

paradigm:
Resolved the problematic comment.

PR #318, opened 2022-09-27
feat: Apply secure RBAC configuration and New Default Admin Group in stage

jpramos123:
Apply secure RBAC configuration to the platform default RBAC group 

Now the default are read only access to all drift resources

This change STAGE environment

EVNT-697
DRFT-1097


sthirugn:
@casey-williams-rh can you please review the Drift rbac permissions that you requested?

coderbydesign (approved):

PR #264, opened 2022-05-04
 feat: Define role permissions in production environment

romanblanco:
https://issues.redhat.com/browse/RHICOMPL-2712

- [x] ~:warning: waiting for https://github.com/RedHatInsights/compliance-backend/pull/1261 to be deployed to prod~
- [x] We need to test configuration in Stage before merging this.
- [x] ~Also, might need to remove remediation, as mentioned in https://github.com/RedHatInsights/rbac-config/pull/285#issuecomment-1148744152.~
- [x] waiting for https://github.com/RedHatInsights/compliance-frontend/pull/1682
- [x] need to communicate the changes (https://access.redhat.com/articles/6963685)

PR #319, opened 2022-09-27
feat: Apply secure RBAC configuration and New Default Admin Group in prod

jpramos123:
Apply secure RBAC configuration to the platform default RBAC group 

Now the default are read only access to all drift resources

First need to validate and merge PR #318 

EVNT-697-prod

### DIFF
--- a/configs/prod/roles/drift.json
+++ b/configs/prod/roles/drift.json
@@ -4,11 +4,26 @@
       "name": "Drift analysis administrator",
       "description": "Perform any available operation against any Drift Analysis resource.",
       "system": true,
-      "platform_default": true,
-      "version": 8,
+      "admin_default": true,
+      "version": 9,
       "access": [
         {
-          "permission": "drift:*:*"
+          "permission": "drift:comparisons:read"
+        },
+        {
+          "permission": "drift:baselines:read"
+        },
+        {
+          "permission": "drift:baselines:write"
+        },
+        {
+          "permission": "drift:historical-system-profiles:read"
+        },
+        {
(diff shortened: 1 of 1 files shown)


configs/prod/roles/drift.json
```diff
@@ -1,14 +1,20 @@
 {
   "roles": [
     {
-      "name": "Drift analysis administrator",
-      "description": "Perform any available operation against any Drift Analysis resource.",
+      "name": "Drift viewer",
```
sthirugn, line 4: I think this is incorrect.

1. The idea is to retain `Drift analysis administrator` and make it `"admin_default": true,` rather than `"platform_default": true,`  see example: https://github.com/RedHatInsights/rbac-config/pull/264/files
2. Perhaps create a separate `Drift viewer` permission like shown in the above link and add all drift related read-only permissions to those.

cc @jeromemarc

configs/prod/roles/drift.json
```diff
@@ -1,14 +1,20 @@
...
-      "version": 8,
+      "version": 9,
       "access": [
         {
-          "permission": "drift:*:*"
+          "permission": "drift:comparisons:read"
+        },
+        {
+          "permission": "drift:baselines:read"
+        },
```
sthirugn, line 15: also add baselines write here

configs/prod/roles/drift.json
jpramos123: @sthirugn I made the changes!

Thanks

sthirugn:
this is pending stage testing.

coderbydesign (approved):

PR #335, opened 2022-10-21
Add Inventory Groups roles

fstavela:
Inventory Administrator role wasn't the default for admins, because it was default for everyone. But, the default user access group is editable and when the Inventory Administrator role is removed from it, the org admin automatically loses the permissions too. That's why I think the Inventory Administrator role should be default for admins too.

_EDIT: In addition to the original purpose of this PR, there are Inventory Groups roles added to the RBAC config. As part of that, there are new, more granular roles created. Now we can remove defaults from "Inventory Administrator" and make "Inventory Hosts Administrator" default instead._

### Links
[JIRA - Add Inventory Groups roles](https://issues.redhat.com/browse/RHCLOUD-23435)

### DIFF
--- a/configs/stage/roles/inventory.json
+++ b/configs/stage/roles/inventory.json
@@ -4,11 +4,70 @@
       "name": "Inventory administrator",
       "description": "Perform any available operation against any Inventory resource.",
       "system": true,
-      "platform_default": true,
-      "version": 2,
+      "platform_default": false,
+      "admin_default": false,
+      "version": 3,
       "access": [
         {
           "permission": "inventory:*:*"
+        }
+      ]
+    },
+    {
+      "name": "Inventory Hosts Administrator",
+      "description": "Be able to read and edit Inventory Hosts data.",
+      "system": true,
+      "platform_default": true,
+      "admin_default": true,
+      "version": 3,
+      "access": [
+        {
+          "permission": "inventory:hosts:write"
+        },
+        {
+          "permission": "inventory:hosts:read"
+        }
+      ]
+    },
+    {
+      "name": "Inventory Hosts Viewer",
+      "description": "Be able to read Inventory Hosts data.",
+      "system": true,
+      "platform_default": false,
+      "admin_default": false,
+      "version": 3,
+      "access": [
+        {
+          "permission": "inventory:hosts:read"
+        }
+      ]
+    },
+    {
+      "name": "Inventory Groups Administrator",
+      "description": "Be able to read and edit Inventory Groups data.",
+      "system": true,
+      "platform_default": false,
+      "admin_default": true,
+      "version": 3,
+      "access": [
+        {
+          "permission": "inventory:groups:write"
+        },
+        {
+          "permission": "inventory:groups:read"
+        }
+      ]
+    },
+    {
+      "name": "Inventory Groups Viewer",
+      "description": "Be able to read Inventory Groups data.",
+      "system": true,
+      "platform_default": false,
+      "admin_default": false,
+      "version": 3,
+      "access": [
+        {
+          "permission": "inventory:groups:read"
         }
       ]
     }
--- a/configs/prod/permissions/inventory.json
+++ b/configs/prod/permissions/inventory.json
@@ -17,5 +17,16 @@
         {
             "verb": "*"
         }
+    ],
+    "groups": [
+        {
+            "verb": "read"
+        },
+        {
+            "verb": "write"
+        },
+        {
+            "verb": "*"
+        }
     ]
 }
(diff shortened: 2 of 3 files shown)


FabriciaDinizRH (approved):

CodeHeeler (approved):
LGTM

kruai (approved):

configs/fedramp-prod/roles/inventory.json
```diff
@@ -5,6 +5,7 @@
       "description": "Perform any available operation against any Inventory resource.",
       "system": true,
       "platform_default": true,
+      "admin_default": true,
```
lpichler, line 8: we don't need to change files fedram-* directories.

configs/stage/roles/inventory.json
```diff
@@ -5,6 +5,7 @@
       "description": "Perform any available operation against any Inventory resource.",
       "system": true,
       "platform_default": true,
+      "admin_default": true,
```
lpichler, line 8: as this role has also `platform_default` set to true - it basically gives all permissions to your app for everyone.
Preferred way is to have more granularity on your permissions. If possible it would be great to have more permissions instead of `"inventory:*:*"` but also have read or write actions or also to have also for resources for format of permissions:

`<application>:<resource>:<action>.` 

or if this is not doable - 
 it would be enough to have also `platform_default` set to false.

Let me know whether is it is feasible for you or if you need more info.
thanks !

lpichler:
@fstavela thanks for PR! I had some comments.

configs/fedramp-prod/roles/inventory.json
fstavela: I've heard that there are plans for fedramp environments to come back, that's why also made the change here. But if this is an issue, I don't have a problem with removing this change.

configs/stage/roles/inventory.json
```diff
@@ -5,6 +5,7 @@
...
       "version": 2,
```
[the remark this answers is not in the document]
fstavela: done, thanks for catching it!

configs/stage/roles/inventory.json
fstavela: For now, Inventory has only one resource type - "hosts". With the coming "Inventory Groups" there will probably be more in the future, but not for now.

The reason why I want to set `true` for both `platform_default` and `admin_default` is mentioned in the PR description. By default, we want all users to be able to do anything in their inventory - read, create, update, delete... hosts, that's why `platform_default` was always set to `true` here. The problem is that when an org admin decides to delete this permission from the default user group, then the admin himself also loses the permissions for interacting with the Inventory. And that's why I would like to set both `platform_default` and `admin_default` to `true`.

configs/fedramp-prod/roles/inventory.json
lpichler: yes maybe you can remove it and they will add it if it will be needed, thanks!

configs/stage/roles/inventory.json
lpichler: I am sorry for big delay!

I was discussing this also with security team and when we give admin access by default with wildcard permission - it is not good security practice.

it is best to define more granular permission that way what `resources` and `actions` are intentional.
is this possible to do it ? 

More info here:https://docs.google.com/document/d/1xfpASXhi6C87dl7r1IVSt3a64yrSBqgMgp8YtVvtTa4/edit

thanks!

configs/fedramp-prod/roles/inventory.json
fstavela: Is this comment still valid? Should I remove roles/permissions from fedramp, or should I keep it in the PR?

kruai (approved):

configs/stage/roles/inventory.json
lpichler: yes it looks great! thank you for doing those changes!

configs/fedramp-prod/roles/inventory.json
fstavela: Done

CodeHeeler (approved):
LGTM, thanks for updating to include groups.

lpichler (approved):
Looks good!

lpichler:
@fstavela can we add this to stage for testing ?

fstavela:
> @fstavela can we add this to stage for testing ?

@lpichler Yes, thank you!

lpichler:
@fstavela permission was added to stage, 
can you check it ? Can we add it to production ?

fstavela:
> 

Update: This change was added to production and works fine.

lpichler:
> Update: This change was added to production and works fine.

Thanks @fstavela !

PR #336, opened 2022-10-27
Add basic roles/permissions for content-sources app

Andrewgdewar:
These permissions are needed to enable permission integration on our platform.

Big thanks to @coderbydesign  for his help.

### DIFF
--- a/configs/stage/permissions/content-sources.json
+++ b/configs/stage/permissions/content-sources.json
@@ -1,0 +1,15 @@
+{
+    "repositories": [
+        {
+            "verb": "read"
+        },
+        {
+            "verb": "write"
+        }
+    ],
+    "*": [
+        {
+            "verb": "*"
+        }
+    ]
+}
--- a/configs/stage/roles/content-sources.json
+++ b/configs/stage/roles/content-sources.json
@@ -1,0 +1,32 @@
+{
+    "roles": [
+      {
+        "name": "Repositories administrator",
+        "description": "Perform any available operation against any repositories resource.",
+        "system": true,
+        "platform_default": false,
+        "admin_default": true,
+        "version": 2,
+        "access": [
+          {
+            "permission": "content-sources:repositories:read"
+          },
+          {
+            "permission": "content-sources:repositories:write"
+          }
+        ]
+      },
+      {
+        "name": "Repositories viewer",
+        "description": "Perform read only operations against repositories resources.",
+        "system": true,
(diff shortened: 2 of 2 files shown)


jlsherrill (approved):
As team lead on content-sources, i approve :)

configs/stage/permissions/content-sources.json
```diff
@@ -0,0 +1,15 @@
+{
+    "policies": [
```
jlsherrill, line 2: actually, this should likely be 'repositories' or 'content-sources'?

Andrewgdewar: oh derp derp derp hard...

Andrewgdewar: Shhh.... you saw nothing Justin Snow...

configs/stage/roles/content-sources.json
```diff
@@ -0,0 +1,29 @@
+{
+  "roles": [
+    {
+      "name": "Content Sources administrator",
+      "description": "Perform any available operation against any Content Source resource.",
```
jlsherrill, line 5: since the app is branded 'Repositories', should we really put that in the name & description?

configs/stage/roles/content-sources.json
```diff
@@ -0,0 +1,32 @@
...
+            "permission": "content-sources:repositories:write"
+          }
+        ]
+      },
+      {
+        "name": "Repositories viewer",
+        "description": "Perform read only operations against repositories resources.",
+        "system": true,
+        "platform_default": true,
+        "version": 2,
```
jlsherrill, line 24: should this be version 1?

Andrewgdewar: no, as I understand it version 2 ensures that it propagates to all current users if I understand [this documentation](https://github.com/RedHatInsights/rbac-config#canned-roles) correctly.

Andrewgdewar: @coderbydesign ^^

syncrou:
cc - @lpichler @lindgrenj6

configs/stage/roles/content-sources.json
coderbydesign: This should be fine as is. We need to update the README to reflect that new roles can now start at 1, but starting at 2 works as well! :)

coderbydesign (approved):

PR #338, opened 2022-11-07
use extended patch permissions in prod

psegedy:
Promote changes from https://github.com/RedHatInsights/rbac-config/pull/326 to prod

### DIFF
--- a/configs/prod/roles/patch.json
+++ b/configs/prod/roles/patch.json
@@ -4,11 +4,34 @@
       "name": "Patch administrator",
       "description": "Perform any available operation against any Patch resource.",
       "system": true,
-      "platform_default": true,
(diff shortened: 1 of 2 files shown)


configs/prod/roles/patch.json
```diff
@@ -4,12 +4,35 @@
...
+        {
+          "permission": "remediations:*:write"
+        }
+      ]
+    },
+    {
+      "name": "Patch viewer",
+      "description": "Read any Patch resource.",
+      "system": true,
+      "platform_default": true,
```
lpichler, line 29: @psegedy just for confirmation - `platform_default: true` means that role is given to everyone by default,

is this correct ?

PR #340, opened 2022-11-24
Bump versions of Patch Roles in stage

lpichler:
we need to bump versions of patch roles to run seeds.

https://issues.redhat.com/browse/RHCLOUD-23137


lpichler:
I am this as this was created: https://github.com/RedHatInsights/rbac-config/pull/341

PR #356, opened 2023-02-07
Add Default access OCM groups

paradigm:
OCM needs the Default access group to include two roles, `OCM Cluster Provisioner` and `OCM Cluster Viewer`.


lpichler (approved):

lpichler:
Changes were added to stage and production environment.

PR #367, opened 2023-04-17
fix(rbac): Remove wildcard permission from admin role

jpramos123:
Ref: POL-726


abaiken (commented):
I think the version needs to be bumped

jpramos123:
> I think the version needs to be bumped

Done!

Thanks

abaiken (approved):

PR #370, opened 2023-04-21
Seed OCM idp and machine pool roles in stage

rawsyntax:
JIRA SDA-6540

### DIFF
--- a/configs/stage/roles/ocm.json
+++ b/configs/stage/roles/ocm.json
@@ -13,6 +13,30 @@
       }
     },
     {
+      "name": "OCM Idp Editor",
+      "display_name": "OCM Idp Editor",
+      "description": "Grants permission to edit idps",
+      "system": true,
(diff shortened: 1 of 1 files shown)


rawsyntax:
@coderbydesign @wcmitchell @paradigm please take a look

configs/stage/roles/ocm.json
```diff
@@ -12,6 +12,30 @@
         "tenant": "ocm"
       }
     },
+    {
+      "name": "OCM Idp Editor",
+      "display_name": "OCM Idp Editor",
+      "description": "Grants permission to edit idps",
+      "system": true,
+      "platform_default": false,
+      "version": 4,
```
lpichler, line 21: This will work but is there reason to start version with 4 ? (version 1 is usually used for new roles)

lpichler (approved):

PR #406, opened 2023-08-08
Update platform default and promote provisioning roles/perms to prod

lzap:
This promotes provisioning perms/roles, which are in stage for quite some time now, to prod.

No changes were made, these are copies of stage files.

Is it possible to get to production by Aug 13th? Thanks.

### DIFF
--- a/configs/prod/permissions/provisioning.json
+++ b/configs/prod/permissions/provisioning.json
@@ -1,0 +1,85 @@
+{
+    "source": [
+        {
+            "verb": "*"
+        },
+        {
+            "verb": "read"
+        }
+    ],
+    "pubkey": [
+        {
+            "verb": "*"
+        },
+        {
+            "verb": "read"
+        },
+        {
+            "verb": "write",
+            "requires": [
+                "read"
+            ]
+        }
+    ],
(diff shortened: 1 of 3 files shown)


lpichler:
@lzap We are going to do it on Thursday.

lzap:
I added one more one-liner: we want the Launch Viewer to be in the Default Access Group, so I added `platform_default` flag on stage and then in another commit I promote both files to prod.

lzap:
@lpichler we are pushing to prod today, ready to flip the RBAC feature on in that environment when you say so. Thanks.

lpichler:
> @lpichler we are pushing to prod today, ready to flip the RBAC feature on in that environment when you say so. Thanks.

@lzap we are going to do it in 2-3 hours approx. But in this case we will do stage first.

abaiken (approved):

lpichler:
@lzap this is in prod, can you check it please ?

lzap:
Well my account on prod does not have access to IdM. Need to figure out how to check that.

lzap:
A colleague checked roles, they are there. Thank you.

PR #407, opened 2023-08-08
Add Hybrid Committed Spend roles and permissions in stage

lorenzobenvenuti:
Hi,

console.redhat.com is accessing one of our APIs (billing.api.redhat.com) in order to retrieve Hybrid Commited Spend reports. This PR adds a role and a permission (in stage only, for now) that will be used to integrate our application with Console RBAC.

For reference: https://issues.redhat.com/browse/ITEAIFIN-2299

Thanks,

lorenzo

### DIFF
--- a/configs/stage/roles/hybrid-committed-spend.json
+++ b/configs/stage/roles/hybrid-committed-spend.json
@@ -1,0 +1,17 @@
+{
+  "roles": [
+    {
+      "name": "Hybrid Committed Spend viewer",
+      "description": "View any Hybrid Committed Spend report.",
+      "system": false,
+      "admin_default": false,
+      "platform_default": false,
+      "version": 1,
+      "access": [
+        {
+          "permission": "hybrid-committed-spend:reports:read"
+        }
+      ]
+    }
+  ]
+}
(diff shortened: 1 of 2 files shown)


configs/stage/roles/hybrid-committed-spend.json
```diff
@@ -0,0 +1,17 @@
+{
+  "roles": [
+    {
+      "name": "Hybrid Committed Spend viewer",
+      "description": "View any Hybrid Committed Spend report.",
+      "system": false,
```
lpichler, line 6: @lorenzobenvenuti do you want non-system role ? 
- it could be deleted and updated - preference here is to have system role which is seeded.

Note: I found that we are setting to [True](https://github.com/RedHatInsights/insights-rbac/blob/master/rbac/management/role/definer.py#L71) by default during seeding. So maybe it will be better to set it to true.

lorenzobenvenuti: Thanks @lpichler , I overlooked that field. Updated to `true`.

Thanks,

lorenzo

lpichler (approved):

lpichler:
https://issues.redhat.com/browse/RHCLOUD-27578

PR #419, opened 2023-08-16
feat(ESSNTL-5163): Adding Staleness and Culling

fellipeh:
Adding Account Staleness and Culling permissions

### DIFF
--- a/configs/stage/roles/inventory.json
+++ b/configs/stage/roles/inventory.json
@@ -70,6 +70,35 @@
           "permission": "inventory:groups:read"
         }
       ]
+    },
+    {
+      "name": "Account Staleness and Culling Administrator",
+      "description": "Be able to read and edit Account Staleness and Culling data.",
+      "system": true,
+      "platform_default": false,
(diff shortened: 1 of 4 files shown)


configs/stage/roles/inventory.json
```diff
@@ -70,6 +70,35 @@
...
+        {
+          "permission": "inventory:staleness:read"
+        }
+      ]
+    },
+    {
+      "name": "Account Staleness and Culling Viewer",
+      "description": "Be able to read Account Staleness and Culling data.",
+      "system": true,
+      "platform_default": false,
```
jpramos123, line 94: I think this should be `true`, as peer this statement:

`Regular users will have the “Inventory staleness and culling viewer” role by default`

See: https://consoledot.pages.redhat.com/docs/dev/services/rbac.html#_define_platform_default_roles
```suggestion
      "platform_default": true,
```

jpramos123 (changes requested):

jpramos123 (approved):

PR #431, opened 2023-09-18
Change inventory.json, replace "Culling" with "Deletion"

spank50:
@abaiken - Updates to inventory.json
Replace Culling with Deletion and in each changed description,
change version 3 to version 4

### DIFF
--- a/configs/prod/roles/inventory.json
+++ b/configs/prod/roles/inventory.json
@@ -72,12 +72,12 @@
       ]
     },
     {
-      "name": "Account Staleness and Culling Administrator",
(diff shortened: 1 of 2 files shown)


lpichler:
@spank50 this is replacement https://github.com/RedHatInsights/rbac-config/pull/434 

let us know if you questions - name is changed by `display_name` .

spank50:
@lpichler  -I saw the guidelines that name changes require "display_name" update -- changing the "name" apparently creates a new role, not what we want. Thanks.

PR #441, opened 2023-10-09
feat(ESSNTL-5486): Add new staleness application, permissions and roles

jpramos123:


abaiken (commented):
Can you break this out into two separate PRs - one for stage & one for prod so that we can test the stage changes/not be blocked on promoting other things to prod at the same time 😁 TIA!

jpramos123:
> Can you break this out into two separate PRs - one for stage & one for prod so that we can test the stage changes/not be blocked on promoting other things to prod at the same time 😁 TIA!

Done! Below are both PRs:

Stage: https://github.com/RedHatInsights/rbac-config/pull/442
Prod: https://github.com/RedHatInsights/rbac-config/pull/443

Thanks

PR #452, opened 2023-10-27
TEAMNADO-5688 - Update RHC Viewer to RHC User and change permission in Stage

jmbuscemi:
Per https://issues.redhat.com/browse/TEAMNADO-5688

This changes the `RHC Viewer` role to `RHC User` and updates permission to add `config-manager:activation_keys:write`

This change is only for Stage. After testing is complete in Stage we'll open up another PR for Prod.

### DIFF
--- a/configs/stage/roles/config-manager.json
+++ b/configs/stage/roles/config-manager.json
@@ -39,7 +39,7 @@
       ]
     },
     {
-      "name": "RHC User",
+      "name": "RHC Viewer",
       "display_name": "RHC user",
       "description": "Can view the current configurations on RHC manager and write to activation keys",
       "system": true,
--- a/configs/stage/roles/config-manager.json
+++ b/configs/stage/roles/config-manager.json
@@ -39,15 +39,15 @@
       ]
     },
     {
-      "name": "RHC Viewer",
-      "display_name": "RHC viewer",
-      "description": "Can view the current configurations on RHC manager",
+      "name": "RHC User",
+      "display_name": "RHC user",
+      "description": "Can view the current configurations on RHC manager and write to activation keys",
       "system": true,
       "platform_default": true,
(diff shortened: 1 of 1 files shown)


configs/stage/roles/config-manager.json
```diff
@@ -39,16 +39,19 @@
       ]
     },
     {
-      "name": "RHC Viewer",
-      "display_name": "RHC viewer",
-      "description": "Can view the current configurations on RHC manager",
+      "name": "RHC User",
```
abaiken, line 42: If you are trying to change the name from `RHC Viewer` to `RHC User` can you change this PR to just update the `display_name` to the name that you want. Can you also update the version number: 

> Update roles

> When you update the role, please change the version number of the role for the service to pick up new features. If you need to update the name of a role, add or update the display_name field in the role config, as this is the the database value displayed in the UI. If you do not have this set, RBAC defaults display_name to the value of the name field in your role config.

jmbuscemi: I'm not sure I follow. We need to change the name *and* the associated permissions.

Also, for consistency's sake, wouldn't I change both `name` and `display_name` ?

abaiken: Once a role has been seeded/exists in RBAC's database, if you update the `name` field and the `display_name` field above it is going to create a new role in our database. If you instead just update the `display_name` - it will update the name that is shown to Users without creating a new role for RBAC. It is fine to also add the permissions. I'm just asking to not update the `name` field so that a new role is not created.

jmbuscemi: Ahh...I see. Thanks for explaining that.

It's not my favorite thing in the world, but that's how the system is built :grimacing:

abaiken (approved):

PR #478, opened 2023-12-19
RHCLOUD-29986 | feature: make Notifications viewer as a default role

MikelAlejoBR:
We want to encourage users to interact with Notifications more by at least allowing them to see what is configured for their tenants.

## Jira ticket
[[RHCLOUD-29986]](https://issues.redhat.com/browse/RHCLOUD-29986)

### DIFF
--- a/configs/prod/roles/notifications.json
+++ b/configs/prod/roles/notifications.json
@@ -20,8 +20,8 @@
       "name": "Notifications viewer",
       "description": "Read only access to notifications and integrations applications.",
       "system": true,
-      "platform_default": false,
-      "version": 2,
+      "platform_default": true,
+      "version": 3,
       "access": [
         {
           "permission": "notifications:notifications:read"
--- a/configs/stage/roles/notifications.json
+++ b/configs/stage/roles/notifications.json
@@ -20,8 +20,8 @@
       "name": "Notifications viewer",
       "description": "Read only access to notifications and integrations applications.",
       "system": true,
-      "platform_default": false,
-      "version": 2,
+      "platform_default": true,
+      "version": 3,
       "access": [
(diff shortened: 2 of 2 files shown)


astrozzc (approved):
LGTM!

lpichler:
NOTE: this probably produces lot of notification events. (as expected)

MikelAlejoBR:
Thanks for the heads up, @lpichler!

configs/stage/roles/notifications.json
```diff
@@ -20,7 +20,7 @@
       "name": "Notifications viewer",
       "description": "Read only access to notifications and integrations applications.",
       "system": true,
-      "platform_default": false,
+      "platform_default": true,
       "version": 2,
```
abaiken, line 24: I think we need to bump the version

configs/prod/roles/notifications.json
```diff
@@ -20,7 +20,7 @@
       "name": "Notifications viewer",
       "description": "Read only access to notifications and integrations applications.",
       "system": true,
-      "platform_default": false,
+      "platform_default": true,
       "version": 2,
```
abaiken, line 24: I think we need to bump the version

abaiken (changes requested):
I can do a stage release on this early next week and then we can try a prod release after 🥇

configs/prod/roles/notifications.json
astrozzc: Oh! Right! Good catch!

MikelAlejoBR: Changed as well, thanks!

abaiken (approved):

PR #482, opened 2024-02-15
HMS-3520: add idmsvc permissions and default roles

avisiedo:
This change provision and the configuration of the permissions and default roles that will manage the idmsvc microservice.

This provide the configuration for the stage environment.

### DIFF
--- a/configs/stage/permissions/idmsvc.json
+++ b/configs/stage/permissions/idmsvc.json
@@ -1,0 +1,29 @@
+{
+    "token": [
+        {
+            "verb": "create"
+        }
+    ],
+    "domains": [
+        {
+            "verb": "list"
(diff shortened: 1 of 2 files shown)


lpichler (approved):
This looks perfect 👍 💯 

I'll ask @abaiken for second check. 

When do you need it in stage ?

avisiedo:
thanks @lpichler , in short for the next week.

I think the cadence on Tuesday is ok; we will start to code the middleware, and we will need to request to the rbac service in stage for checking the expected responses so the middleware integrate well with it; if I am not missing something!

abaiken:
@avisiedo This should now be in stage - let me know if you have any issues!

PR #492, opened 2024-04-09
Update to sources administrator role name (RHCLOUD-31937)

rabbott01:
Updating the sources administrator role name to "Cloud Administrator" per https://issues.redhat.com/browse/RHCLOUD-31937

### DIFF
--- a/configs/prod/roles/sources.json
+++ b/configs/prod/roles/sources.json
@@ -1,7 +1,8 @@
 {
   "roles": [
     {
-      "name": "Cloud Administrator",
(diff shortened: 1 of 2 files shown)


petracihalova (changes requested):
in this case we need to add the `display_name` instead, the `name` should stay same

rabbott01:
Thanks @petracihalova - updated and pushed the commits.

petracihalova (approved):

lpichler (approved):

PR #497, opened 2024-06-17
Update subscriptions.json

rabbott01:
Updating "Subscription user" to "Subscription viewer" per RHCLOUD-33136

### DIFF
--- a/configs/prod/roles/subscriptions.json
+++ b/configs/prod/roles/subscriptions.json
@@ -15,7 +15,7 @@
       ]
     },
     {
-      "name": "Subscriptions viewer",
+      "name": "Subscriptions user",
       "display_name": "Subscription viewer",
       "description": "View any Subscriptions resource.",
       "system": true,
--- a/configs/prod/roles/subscriptions.json
+++ b/configs/prod/roles/subscriptions.json
@@ -15,7 +15,7 @@
       ]
     },
     {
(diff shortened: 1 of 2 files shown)


configs/prod/roles/subscriptions.json
```diff
@@ -15,11 +15,11 @@
       ]
     },
     {
-      "name": "Subscriptions user",
+      "name": "Subscriptions viewer",
```
petracihalova, line 18: please let the "name" same and add the "display_name" field as we did here https://github.com/RedHatInsights/rbac-config/pull/492

petracihalova (changes requested):

petracihalova:
> Thanks @petracihalova - I added the "display_name" to Stage/Prod configs.

awesome, thank you for adding "display_name" please can you change back the "name" field?

configs/prod/roles/subscriptions.json
```diff
@@ -16,10 +16,11 @@
     },
     {
       "name": "Subscriptions user",
+      "display_name": "Subscription viewer",
```
petracihalova, line 19:
```suggestion
      "display_name": "Subscriptions viewer",
```

lpichler:
@rabbott01 if you can squash commit, it would be awesome!

maybe you can increment version by 1 (no by 2)

PR #499, opened 2024-06-20
Update subscriptions.json

rabbott01:
Updated version for Subscription User to 5 (previously went from 4 to 6)


rabbott01:
@lpichler updated the versions to jump by 1 v. 2 from the previous PR

PR #509, opened 2024-07-11
RHINENG-10498(stage): Adding Advisor read rule

fellipeh:
Adding `advisor` read only permission and rule

Closes: RHINENG-10498

PR #510, opened 2024-07-11
RHINENG-10498(prod): Adding Advisor read rule

fellipeh:
Adding `advisor` read only permission and rule


configs/prod/roles/insights.json
```diff
@@ -15,6 +15,20 @@
...
       ]
+    },
+    {
+          "name": "Advisor Viewer",
+          "display_name": "RHEL Advisor viewer",
+          "description": "Be able to read Advisor data.",
+          "system": true,
+          "platform_default": false,
+          "admin_default": false,
+          "version": 10,
```
petracihalova, line 26: please version should be 1 😘

configs/prod/roles/insights.json
```diff
@@ -15,6 +15,20 @@
           "permission": "inventory:*:read"
         }
       ]
+    },
+    {
+          "name": "Advisor Viewer",
```
petracihalova, line 20: 👍  I'm curious so this permission and role will only exist on prod?

petracihalova: now I understand 😂  https://github.com/RedHatInsights/rbac-config/pull/509

fellipeh: yeah.. I split the change, because long time ago was told me to make separeted PR for prod and stage, so, if there's an issue on deployment It's will be more easy to revert only one..

petracihalova (approved):

PR #534, opened 2024-09-12
Upate RBAC role descriptions for stage

spank50:
RHCLOUD-35164

PR #536, opened 2024-09-13
fix(HMS-4719): update role name and description

avisiedo:
Update the role name and description for the idmsvc roles.

https://issues.redhat.com/browse/HMS-4719


configs/stage/roles/idmsvc.json
```diff
@@ -1,12 +1,12 @@
...
     {
-      "name": "Domains administrator",
-      "description": "Perform any available operation against any domain resource.",
+      "name": "Directory and Domain Services administrator",
+      "description": "Create, read, update and delete Directory and Domains Services registrations.",
       "system": true,
       "platform_default": false,
       "admin_default": true,
-      "version": 1,
+      "version": 2,
```
avisiedo, line 9: Do we need to increase the version? unsure if it is necessary to get the name and description updated in the different environments; please if not needed, let us know and we update the PR. Many thanks!

lpichler: version  needs to be increased and name never needs to be changed - because we are search db object by name. You need to change or  `display_name` instead of them to see new name in UI

lpichler: yes both option are correct. 

In case you will with 2. : 
- change name and don't change versions number.
- please create ticket for us stating name of roles which needs to be deleted
-  there is effort from tech. write to description updates - https://github.com/RedHatInsights/rbac-config/pull/534 - can you work with @spank50 to clarify descriptions for you ?

spank50: @avisiedo - thanks for your attention to the Description field.
This is what I know as a limited contributor:
"name" is shown if there is no "display_name"
If "name" and "display_name" exist, "display_name" is shown. This allows role name updates without creating a new role.
"version" is incremented to flag that there is a change. Otherwise, the change is not published.
"description" fields display in the UI and are the source for the [predefined roles reference](url) chapter

frasertweedale: The new content seems correct to me.

spank50: LGTM

spank50 (approved):

spank50 (approved):
LGTM

lpichler (approved):

PR #546, opened 2024-09-24
[RHCLOUD-35333] add the system role Vulnerability viewer into Default access group

petracihalova:
this change will not affect customers because right now the **Vulnerability administrator** is `platform_default=true`

the current state of Vulnerability system roles in production:
- **Vulnerability administrator** is part of **Default Access** group (with `platform_default=true` flag)
- **Vulnerability viewer** is not part of default groups (without a flag)

the expected state:
- **Vulnerability administrator** is part of **Default Admin Access** group (with `admin_default=true`)
- **Vulnerability viewer** is part of **Default Access** group (with `platform_default=true`)

this PR will fix only the **Vulnerability viewer** role

### DIFF
--- a/configs/prod/roles/vulnerability.json
+++ b/configs/prod/roles/vulnerability.json
@@ -22,8 +22,9 @@
       "name": "Vulnerability viewer",
       "description": "Perform read operations on any Vulnerability resource.",
       "system": true,
-      "platform_default": false,
-      "version": 4,
+      "platform_default": true,
+      "admin_default": false,
+      "version": 5,
       "access": [
         {
           "permission": "vulnerability:vulnerability_results:read"


configs/prod/roles/vulnerability.json
```diff
@@ -22,7 +22,8 @@
       "name": "Vulnerability viewer",
       "description": "Perform read operations on any Vulnerability resource.",
       "system": true,
-      "platform_default": false,
+      "platform_default": true,
+      "admin_default": false,
       "version": 4,
```
astrozzc, line 27: The version has to be updated

petracihalova: yes that is big true thank you I totally missed that

petracihalova: fixed

astrozzc (approved):
LGTM!

PR #569, opened 2024-10-23
RHCLOUD-35813 | feature: remaining permissions for Notifications

MikelAlejoBR:
These temporary permissions will help us deal with the system data like bundles, applications and event types, and the tenant-specific data like behavior groups.

In the future we will refactor the system data calls in the Notifications code to comply with the Kessel style, by querying for the specific elements instead of querying for the broad workspace permissions.

## Jira ticket
[[RHCLOUD-35813]](https://issues.redhat.com/browse/RHCLOUD-35813)


lpichler:
@lennysgarage @wscalf do we need also changes in original RBAC configs ? or just to do changes in original RBAC configs (I guess it will generate changes here ? )?

wscalf:
@lpichler We don't need to update the original rbac configs for this one, no. All the added v2 permissions are based on existing v1 permissions (ex: `app:'notifications', resource:'notifications', verb:'read'` is the equivalent of notifications:notifications:read which is already in the RBAC configs here: https://github.com/RedHatInsights/rbac-config/blob/master/configs/prod/permissions/notifications.json#L4), _but_, if there were new v1 permissions, those would need to be added to the rbac configs as well, otherwise they wouldn't be grantable in the v1 interface.

wscalf (approved):

lpichler (approved):

PR #581, opened 2024-11-15
[RHCLOUD-36286] JSON schema validation fix

petracihalova:
jira: [RHCLOUD-36286](https://issues.redhat.com/browse/RHCLOUD-36286)

Fix for the JSON schema validation
- fix path refs to the json files
  - path to files before `./configs/.*/permissions/.*json`
  - and now `./configs/*/permissions/*.json`
- add additional settings
  - `print-valid-files: true` for listing validated files as an output
  - `strict: false` to not use the strict validation (because the validator is not able to correctly check the required fields defined via 'oneOf')
- add missing permission's verb (with the broken validator we didn't noticed that permission with new verb was added)
- remove duplicity in the `display_name` for `Organization Staleness and Deletion Viewer` role (with the broken validator we didn't noticed that role was added/updated with this duplicity)
- remove `minProperties` and `additionalProperties` definition for `resourceDefinitions` (it can be used only for `type=object` and resourceDefinitions is `type=array`)


abaiken (approved):

lpichler:
great findings @petracihalova !

PR #591, opened 2024-12-02
Add temporary default role for testing the notifications sent post migration

abaiken:
Add a temporary role to stage that is a platform default role to test the number of tenants that are notified


alechenninger (approved):
🚢

coderbydesign (approved):
🌴

PR #593, opened 2025-02-26
Add draft KSL files for HBI

wscalf:
Draft/speculative schema changes for HBI and remediations

Assumptions:
* Host-based service providers will not initially define schemas for their assets- rather, they check permissions against and list _hosts_ by application-specific permission
* Application-specific permissions only apply to hosts that are also viewable (this may not be necessary with more fine-grained role bindings in v2 but should reproduce current behavior with coarse-grained bindings migrated from v1)
* With inventory groups becoming workspaces, HBI is no longer responsible for managing them

Draft solution:
* Transfer inventory:groups:* permissions to RBAC workspaces
* Add host type as a workspaced resource that represents the inventory:hosts:* permissions
* Add expose_host_permission extension that service providers can use to expose a workspace permission via hosts


merlante (approved):
LGTM -- starting point to iterate on. :)

lennysgarage (approved):
LGTM

PR #607, opened 2025-05-15
Define config-manager:profile permissions in stage

jharting:
https://issues.redhat.com/browse/RHINENG-16029
https://issues.redhat.com/browse/RHINENG-8763 & 
its sub-task: https://issues.redhat.com/browse/RHINENG-18274

### DIFF
--- a/configs/stage/permissions/config-manager.json
+++ b/configs/stage/permissions/config-manager.json
@@ -9,5 +9,16 @@
         {
             "verb": "*"
         }
+    ],
+    "profile": [
+        {
+            "verb": "read"
+        },
+        {
+            "verb": "write"
+        },
+        {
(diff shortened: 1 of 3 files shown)


petracihalova:
hi @jharting please add the new permission into config-manager permissions too https://github.com/RedHatInsights/rbac-config/blob/master/configs/stage/permissions/config-manager.json

petracihalova:
@astrozzc please can you help here with review? thank you ❤️

jharting:
@astrozzc @petracihalova @wscalf this should be resolved now. I migrated config_manager:permissions also and added config_manager to the list of migrated apps so that there is no longer a conflict before the new/old profile permission definitions

PR #608, opened 2025-05-15
RHCLOUD-39124: Update hbi schema files

Adam0Brien:
Closes: RHCLOUD-39124

Due to descoping, only the staleness permissions (at the workspace level) are needed for naive implementation.

### DIFF
--- a/configs/prod/schemas/src/hbi.ksl
+++ b/configs/prod/schemas/src/hbi.ksl
@@ -3,6 +3,9 @@ namespace hbi
 
 import rbac
 
+@rbac.add_v1_based_permission(app:'staleness', resource:'staleness', verb:'read', v2_perm:'staleness_staleness_view');
+@rbac.add_v1_based_permission(app:'staleness', resource:'staleness', verb:'write', v2_perm:'staleness_staleness_update');
+
 public type host {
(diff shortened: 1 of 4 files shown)


configs/prod/schemas/src/hbi.ksl
```diff
@@ -3,6 +3,9 @@ namespace hbi
 
 import rbac
 
+@rbac.add_v1_based_permission(app:'staleness', resource:'staleness', verb:'read', v2_perm:'staleness_read_staleness');
```
wscalf, line 6: I remembered after the call that the other permissions (and the patterns in the HBI proof of concept) are formatted '${application}_${resource}_${verb}' like the role permissions but use alternate verbs (read -> view, write -> update)

So we probably want staleness_staleness_view and staleness_staleness_update here.

configs/stage/schemas/src/hbi.ksl
```diff
@@ -3,6 +3,9 @@ namespace hbi
 
 import rbac
 
+@rbac.add_v1_based_permission(app:'staleness', resource:'staleness', verb:'read', v2_perm:'staleness_read_staleness');
```
wscalf, line 6: Also here

wscalf (approved):

PR #615, opened 2025-06-12
cost mgmt removal of depricated oci roles

lcouzens:
Clean up of old OCI roles that are no longer needed.


petracihalova:
@lcouzens hi Luke, please can you split your changes for stage and prod into 2 separate PRs? and please for role updates we need to raise the "version" field too

PR #647, opened 2025-07-31
RHINENG-19334 Kessel permissions for Compliance

romanblanco:
https://project-kessel.github.io/docs/building-with-kessel/how-to/migrate-from-rbac-v1-to-v2/#model-the-permissions

### DIFF
--- a/configs/stage/schemas/src/compliance.ksl
+++ b/configs/stage/schemas/src/compliance.ksl
@@ -9,13 +9,9 @@ import rbac
 @rbac.add_v1_based_permission(app:'compliance', resource:'policy', verb:'write', v2_perm:'compliance_policy_edit');
 @rbac.add_v1_based_permission(app:'compliance', resource:'policy', verb:'delete', v2_perm:'compliance_policy_remove');
 
-// This permission is not used by compliance backend
-// But frontend uses it here: https://github.com/RedHatInsights/compliance-frontend/blob/cd020f48920f51f0422256e7d91beada13e5d5d4/src/Routes.js#L103
-// And it is part of the roles: https://github.com/search?q=org%3ARedHatInsights+compliance%3Apolicy%3Aupdate&type=code
+// These 2 permissions are not used by compliance backend - they were used by the frontend but they updated to align with the backend
+// https://github.com/RedHatInsights/compliance-frontend/pull/2654/files
 @rbac.add_v1only_permission(perm:'compliance_policy_update');
-
-// https://github.com/RedHatInsights/compliance-frontend/blob/cd020f48920f51f0422256e7d91beada13e5d5d4/src/Routes.js#L38
-// https://github.com/search?q=org%3ARedHatInsights%20compliance%3Areport%3Adelete&type=code
 @rbac.add_v1only_permission(perm:'compliance_report_delete');
 
 // Report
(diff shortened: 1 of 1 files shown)


wscalf:
Oh! I see what's going on. Two things, actually.

The first is declaring 'all' permissions. Those aren't really meant to be declared or checked explicitly- they're meant to be an internal implementation detail for how a permission can be granted, while the permissions are meant to be concrete things a user can do. The reason you're getting a conflict is the `rbac.add_v1_based_permission` extension is _also_ trying to add them. See: https://github.com/romanblanco/rbac-config/blob/kessel-compliance-schema/configs/stage/schemas/src/rbac.ksl#L46

For these, I'd first ask if you have logic that depends on wildcards currently, and if so if it could be thought of in terms of an operation against a thing- that would fit our model better.

The other conflict you'd hit is here: https://github.com/romanblanco/rbac-config/blob/kessel-compliance-schema/configs/stage/schemas/src/compliance.ksl#L25

It's creating compliance_report_view twice, once as a v1-based permission and once as a contingent permission. You probably meant:
```
@rbac.add_v1_based_permission(app:'compliance', resource:'report', verb:'read', v2_perm:'compliance_report_view_assigned');
@rbac.add_contingent_permission(first: 'inventory_host_view', second: 'compliance_report_view_assigned', contingent: 'compliance_report_view');
```

configs/stage/schemas/src/compliance.ksl
```diff
@@ -0,0 +1,29 @@
...
+@rbac.add_v1_based_permission(app:'compliance', resource:'policy', verb:'delete', v2_perm:'compliance_policy_remove');
+
+// This permission is not used by compliance backend
+// But frontend uses it here: https://github.com/RedHatInsights/compliance-frontend/blob/cd020f48920f51f0422256e7d91beada13e5d5d4/src/Routes.js#L103
+// And it is part of the roles: https://github.com/search?q=org%3ARedHatInsights+compliance%3Apolicy%3Aupdate&type=code
+@rbac.add_v1only_permission(perm:'compliance_policy_update');
+
+// https://github.com/RedHatInsights/compliance-frontend/blob/cd020f48920f51f0422256e7d91beada13e5d5d4/src/Routes.js#L38
+// https://github.com/search?q=org%3ARedHatInsights%20compliance%3Areport%3Adelete&type=code
+@rbac.add_v1only_permission(perm:'compliance_report_delete');
```
wscalf, line 19: I think this is fine for now, but note that v1only permissions are essentially just containers - the permissions here won't be accessible via Kessel until they're mapped to something V2.

configs/stage/schemas/src/compliance.ksl
```diff
@@ -0,0 +1,29 @@
...
+// Policy
+@rbac.add_v1_based_permission(app:'compliance', resource:'policy', verb:'read', v2_perm:'compliance_policy_view');
+@rbac.add_v1_based_permission(app:'compliance', resource:'policy', verb:'create', v2_perm:'compliance_policy_new');
+@rbac.add_v1_based_permission(app:'compliance', resource:'policy', verb:'write', v2_perm:'compliance_policy_edit');
+@rbac.add_v1_based_permission(app:'compliance', resource:'policy', verb:'delete', v2_perm:'compliance_policy_remove');
+
+// This permission is not used by compliance backend
+// But frontend uses it here: https://github.com/RedHatInsights/compliance-frontend/blob/cd020f48920f51f0422256e7d91beada13e5d5d4/src/Routes.js#L103
+// And it is part of the roles: https://github.com/search?q=org%3ARedHatInsights+compliance%3Apolicy%3Aupdate&type=code
+@rbac.add_v1only_permission(perm:'compliance_policy_update');
```
wscalf, line 15: Same point on v1only permissions

wscalf (approved):
Looks fine, just note that points on v1only permissions - this will probably need to be revisited at some point

configs/stage/schemas/src/compliance.ksl
josejulio: These are not used anymore

romanblanco (commented):
@josejulio, LGTM 👍

PR #648, opened 2025-07-31
SWATCH-3830: Add Kessel permission definitions for subscriptions (stage)

lindseyburnett:
- Create subscriptions.ksl in stage environment with v1-based permissions
- Convert existing permissions to Kessel Schema Language format
- Use @rbac.add_v1_based_permission() extension for compatibility
- Maps all subscriptions permissions: reports, manifests, organization, products, cloud_access

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/configs/stage/schemas/src/subscriptions.ksl
+++ b/configs/stage/schemas/src/subscriptions.ksl
@@ -1,0 +1,16 @@
+version 0.1
+namespace subscriptions
+
+import rbac
+
+@rbac.add_v1_based_permission(app:'subscriptions', resource:'report', verb:'read', v2_perm:'subscriptions_report_view');
+@rbac.add_v1_based_permission(app:'subscriptions', resource:'subscription', verb:'read', v2_perm:'subscriptions_subscription_view');
+@rbac.add_v1_based_permission(app:'subscriptions', resource:'subscription', verb:'write', v2_perm:'subscriptions_subscription_edit');
+@rbac.add_v1_based_permission(app:'subscriptions', resource:'manifest', verb:'read', v2_perm:'subscriptions_manifest_view');
+@rbac.add_v1_based_permission(app:'subscriptions', resource:'manifest', verb:'write', v2_perm:'subscriptions_manifest_edit');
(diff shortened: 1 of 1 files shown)


wscalf:
Hey, sorry it took so long to get to this. It looks fine, I just had a question before approving - there are "subscription" permissions in the subscriptions.ksl (ex: `@rbac.add_v1_based_permission(app:'subscriptions', resource:'subscription', verb:'read', v2_perm:'subscriptions_subscription_view');`) that as far as I can tell don't exist in the [permissions.json](https://github.com/RedHatInsights/rbac-config/blob/master/configs/stage/permissions/subscriptions.json) for Subscriptions in RBAC. Those won't be grantable through the RBAC UI. That's potentially fine, they'll be covered by wildcards (`subscriptions:*:read` `subscriptions:*:*`), but I wanted to make sure it was intentional.

lindseyburnett:
@wscalf good catch!  it wasn't intentional.  just pushed a change to remove it.

configs/stage/schemas/src/subscriptions.ksl
```diff
@@ -0,0 +1,14 @@
+version 0.1
+namespace subscriptions
+
+import rbac
+
+@rbac.add_v1_based_permission(app:'subscriptions', resource:'report', verb:'read', v2_perm:'subscriptions_report_view');
```
awood: @lindseyburnett Does this need to be "reports" instead of "report"?  Based on https://github.com/RedHatInsights/rbac-config/blob/master/configs/stage/permissions/subscriptions.json#L2

configs/stage/schemas/src/subscriptions.ksl
```diff
@@ -0,0 +1,14 @@
...
+@rbac.add_v1_based_permission(app:'subscriptions', resource:'manifest', verb:'read', v2_perm:'subscriptions_manifest_view');
```
awood: Likewise with "manifest" and "product"?  Or does RBACv2 not do plurals?

PR #649, opened 2025-08-25
Add content-sources kessel schema to stage

rverdile:


Dugowitch:
Hey @rverdile, I noticed you used plural in resource parameters, [docs](https://improved-adventure-16jykz5.pages.github.io/building-with-kessel/how-to/migrate-from-rbac-v1-to-v2/#model-the-permissions):
> The `resource` parameter should be the **singular** name of the resource type.

Or is that what one should do, when the V1 permission is plural?

rverdile:
> Or is that what one should do, when the V1 permission is plural?

Good question. My understanding is the name of the resource should match the v1 permission. I even get an error building the schema if I make it singular. maybe @wscalf can confirm

wscalf:
Oop, yeah, I think the docs are wrong there. @rverdile is correct - the app/resource/verb parameters are for tying back to RBAC V1 permissions. You can see how that works in [rbac.ksl](https://github.com/RedHatInsights/rbac-config/blob/master/configs/stage/schemas/src/rbac.ksl#L46).

The line `@rbac.add_v1_based_permission(app:'content_sources', resource:'repositories', verb:'read', v2_perm:'content_sources_repository_view');` is essentially saying "if a user is granted content_sources:repositories:read OR content_sources:repositories:* OR content_sources:*:read OR content_sources:*:* OR *:*:* in RBAC, they should have content_sources_repository_view in Kessel"

..And if it anything doesn't match, you'll end up with errors from RBAC at runtime when it tries to populate permissions. One way to check to make sure everything is correct (and that's part of our final PR review process, before the updated schema gets merged) is to build the schema locally (and there's handy README notes for how to do that now!), and then diff the built schema (_private/test-schema/stage-schema.zed) against the current one (configs/stage/schemas/schema.zed) and pay special attention to the role type. Things can be added, but if anything is removed (ie: one of the permission names changed), the parameters to an add_v1_based_permission call are probably mismatched.

PR #650, opened 2025-08-27
Update instructions to build the kessel schema

josejulio:


README.md
```diff
@@ -118,33 +118,23 @@ Building the Kessel schema
 
 Prerequisites
 
-1. Get the `ksl` transpiler tool
+1. Install required libraries by running
```
wscalf, line 121: It's a quibble, but I'd probably use a word like 'tools' here. Or 'toolchain' - it's actually a set (2) of programs that use each other's output to get you to your final output.

We also might want to mention installing a recent go version as one of the prereqs.

On the whole, this looks great! :+1:

PR #652, opened 2025-08-29
Add Kessel schema for Malware

josejulio:


configs/stage/schemas/src/malware.ksl
```diff
@@ -0,0 +1,16 @@
...
+namespace malware
+
+import rbac
+
+// malware-detection:*:read
+@rbac.add_v1_based_permission(app:'malware_detection', resource:'malware', verb:'read', v2_perm:'malware_malware_view_assigned');
+@rbac.add_contingent_permission(first:'inventory_host_view', second:'malware_malware_view_assigned', contingent:'malware_malware_view');
+
+// malware-detection:acknowledgements:write
+@rbac.add_v1_based_permission(app:'malware_detection', resource:'acknowledgements', verb:'write', v2_perm:'malware_malware_acknowledge_assigned');
```
snehagunta, line 11: Do you intend to check this permission off of a host id and if yes, you can use [ this extension ](https://github.com/RedHatInsights/rbac-config/blob/744d8d76ef4a82b93975af9b53638b4e66342368/configs/stage/schemas/src/hbi.ksl#L22) to make it part of the HBI schema and this will enable you to check your permission against the host.

PR #654, opened 2025-09-01
RHINENG-16032 define Kessel permissions for vulnerability

jharting:


configs/stage/schemas/src/vulnerability.ksl
```diff
@@ -0,0 +1,28 @@
+version 0.1
+namespace vulnerability
+
+import rbac
+import hbi
```
lpichler, line 5: Can I ask what is reason for this import ?

PR #795, opened 2026-08-05
[RHCLOUD-49954] Add inventory:views:read/write permissions (in stage)

randomnetcat:
Per [this ticket](https://redhat.atlassian.net/browse/RHCLOUD-49954).

### DIFF
--- a/configs/stage/schemas/src/hbi.ksl
+++ b/configs/stage/schemas/src/hbi.ksl
@@ -6,6 +6,9 @@ import rbac
 @rbac.add_v1_based_permission(app:'staleness', resource:'staleness', verb:'read', v2_perm:'staleness_staleness_view');
 @rbac.add_v1_based_permission(app:'staleness', resource:'staleness', verb:'write', v2_perm:'staleness_staleness_update');
 
+@rbac.add_v1_based_permission(app:'inventory', resource:'views', verb:'read', v2_perm:'inventory_view_view');
+@rbac.add_v1_based_permission(app:'inventory', resource:'views', verb:'write', v2_perm:'inventory_view_update');
+
 public type host {
     private relation workspace: [ExactlyOne rbac.workspace]
     relation tenant: workspace.tenant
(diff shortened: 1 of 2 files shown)


configs/stage/schemas/src/hbi.ksl
```diff
@@ -6,6 +6,9 @@ import rbac
 @rbac.add_v1_based_permission(app:'staleness', resource:'staleness', verb:'read', v2_perm:'staleness_staleness_view');
 @rbac.add_v1_based_permission(app:'staleness', resource:'staleness', verb:'write', v2_perm:'staleness_staleness_update');
 
+@rbac.add_v1_based_permission(app:'inventory', resource:'views', verb:'read', v2_perm:'inventory_view_view');
+@rbac.add_v1_based_permission(app:'inventory', resource:'views', verb:'write', v2_perm:'inventory_view_update');
```
astrozzc, line 10: we don't have v1 permission yet. We might need to create that first.

And maybe worth checking with HBI team, check with aarif if he want to add unified permission - read/write same as v1 or view/update differs in v2.

cc @wscalf for another eye

randomnetcat: > we don't have v1 permission yet. We might need to create that first.

... yes, good point. Sorry.

> And maybe worth checking with HBI team, check with aarif if he want to add unified permission - read/write same as v1 or view/update differs in v2.

`view` and `update` seem to be the status quo for V2; I don't think we want to introduce inconsistencies there? But, sure, I can ask HBI.

randomnetcat: V1 permissions added.

randomnetcat: [HBI is fine with being consistent with the status quo.](https://redhat-internal.slack.com/archives/CQFKM031T/p1785961440279709?thread_ts=1785956427.939909&cid=CQFKM031T)

astrozzc: Cool~, sorry, missed the reviewing. LGTM!

